PR #12, opened 2016-10-13
use errcheck instead of deprecated restype with callable, fixes #11

### LINKED ISSUE #11: SigSegv when using mss on latest ubuntu
General informations:
- OS name: Ubuntu
- OS version: 16.10
- OS architecture: _64 bits_
- Result of the command `python --version`: Python 2.7.12+ and Python 3.5.2+

For GNU/Linux users:
- Display server protocol and version, if known: X.Org X Server 1.18.4
- Desktop Environment: N/A
- Composite Window Manager name and version: qtile 10.6
### Description of the warning/error

trying to use MSS results in a SIGSEGV when calling enum_display_monitors() 
If I run the first lines of the example given in the README

```
>>> from mss import mss
>>> sct = mss(display=b':0')

# Retrieve monitors informations
>>> displays = sct.enum_display_monitors()
```

the program terminates with:
fish: “python” terminated by signal SIGSEGV (Address boundary error)

I've already tried recompiling the C library on my machine but no difference

cycomanic:
__Note:__ do not work on the master branch, only into you own (based on _devel_) or _devel_.

- use errcheck for the validation of the return XRRGetScreenResources, the previous method is deprecated according to ctypes [documentation](https://docs.python.org/2/library/ctypes.html) 
- Fixes #11 

### Code checks

- [X] PEP8 compliant
- [ ] `flake8` passed (bonus)

### DIFF
--- a/mss/linux.py
+++ b/mss/linux.py
@@ -151,7 +151,7 @@ class MSS(MSSBase):
     def _set_restypes(self):
         ''' Functions return type. '''
 
-        def validate(value):
+        def validate(value, func, args):
             ''' Validate the returned value of xrandr.XRRGetScreenResources().
                 We can end on a segfault if not:
                     Xlib:  extension "RANDR" missing on display "...".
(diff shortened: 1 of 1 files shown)


cycomanic:
This fixes #11 for me, however I did not test the case when RANDR is missing, it should work according to the ctypes documentation though.

BoboTiG:
Thanks @cycomanic, it seems to work well.
Do you have a website? I will add you to the CONTRIBUTORS :)

cycomanic:
Thanks, currently I don't have a website, you can use my full name if you want to (Jochen Schroeder)

Issue #14, opened 2016-12-12
Bad multiple screen grab


Archarachne:
I'm working on 10.12.1 MacOS Sierra with 2.0.18 mss version and python 3.5. When today I tried to create multiple screenshots using mss according to documentation.
And with two different screens with resolutions 
{'height': 1200, 'left': 0, 'top': 0, 'width': 1920}, {'height': 768, 'left': -1366, 'top': 0, 'width': 1366}. The bigger screenshot looks fine but the smaller gives a very bad result: http://imgur.com/a/hKrHw. What is the cause of it?

BoboTiG:
Woops, weird. Few questions to try to understand, following [examples](https://python-mss.readthedocs.io/en/latest/examples.html#examples):
1. If you take the screenshot of all monitors, what is the result?
2. What is the result if you take one screenshot per monitor?
3. What is the result if you specify which monitor to capture?

Archarachne:
So following the examples:

1. If both monitors are get a screenshot together the picture looks like the one attached only mixed together.
2. 3 - the screenshot of the bigger, primary screen is always perfect; the other-smaller gets mixed.

BoboTiG:
If you set the little screen as primary one, same results? How is connected the second screen (nothing special on the set-up part, or something helpful?) ?

Archarachne:
It's connected via HDMI and thunderbolt adapter. I'm checking the primary screen right now.

Archarachne:
After switching primary setting between the screens the small one (I own 11 inch MacBook Air with 10.12.1 Sierra) always gives a bad picture while the big one is ok (with bigger resolution).

BoboTiG:
OK, you are using a Thunderbold adaptater, good to know. I will have testing hardware in january, but for now I cannot try something for you :/

Archarachne:
Thanks, will be waiting for an update. 👍 

The funny thing is that the MacBook air screen has got a problem, not the attached monitor. I would expect otherwise. :)

BoboTiG:
Hello,

Finally, I could get a machine and a Thunderbolt adaptater. But all screenshots are good :/
Do you have a clue, something special in your configuration that could interfer? The monitor state i "normal"?

vvarp:
Got hit by the same issue running on MacBook Air 11" - a machine with fairly uncommon native resolution (1366x768)

[...]

PR #15, opened 2017-04-21
Possible fix for BoboTiG/python-mss#14

DavideBecker:
-  Possible fix for #14 
macOS seems to add extra padding in form of black pixels until the width is divisible by 16, this currently can result in a skewed image.
As a solution the image width is rounded up to the next multiple of 16, so we're taking a bigger screenshot than needed and crop it back to its original size afterwards.
Doing it this way doesn't remove important data in case the padding doesn't happen every time or changes in the future.
- An import for `ceil()` has been added in order to easily round up to the next multiple of 16

### Notes
- You should still test this on other machines if possible, since all of this was tested on a single macOS device running Sierra 10.12.4
- I didn't test how these changes impact performance, but it should be negligible and it fixes a pretty important bug
- @redodo helped me figure out much of the proposed solution

- [X] PEP8 compliant

### DIFF
--- a/mss/darwin.py
+++ b/mss/darwin.py
@@ -139,11 +139,11 @@ class MSS(MSSBase):
 
         return self.monitors
 
-    def crop_width(self, image, width_from, width_to, height):
+    def crop_width(self, image, width_to):
         ''' Cut off the pixels from an image buffer at a particular width. '''
         cropped = bytearray()
-        for y in range(height):
-            start = y * width_from * 3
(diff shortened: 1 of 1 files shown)


DavideBecker:
422ac0a fixed an issue where the image wouldn't get cropped back down to its requested size. It also made it PEP8 compliant.

BoboTiG:
Thank you David, nice patch! And performances are still good. Thanks!

PR #46, opened 2018-01-19
Use feature detection instead of version detection

cclauss:
https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

https://travis-ci.org/BoboTiG/python-mss/jobs/330759683#L705


BoboTiG:
Hello,

Thank you! Do you want to add yourself to the contributors file too?

PR #51, opened 2018-03-15
bgra in Snapshot class

eruffaldi:

- Exposes BGRA data so that we avoid the expensive RGB conversion

Is your code right?

- [x] PEP8 compliant


BoboTiG:
Thanks!

PR #52, opened 2018-03-15
Darwin Grabber

eruffaldi:

- Provides preallocated grabber ... small advantage, maybe

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [x] PEP8 compliant


BoboTiG:
Thanks for the PR. Unfortunatly it does complexify the code for no gain. I do not think we can do more on the macOS part.

OTOH If you want to focus on how we compute `RGB` values, it could be very helpful for everyone.

PR #62, opened 2018-08-03
Linux: add an error handler for the XServer to prevent interpreter crash

### LINKED ISSUE #61: [XServer] Library makes program hard crash when monitor is out of bounds
General information:

* OS name: _Ubuntu (`ubuntu-desktop`)_
* OS version: __18.04 LTS__
* OS architecture: _64 bits_
* Resolutions:
    * Monitor 1: _1920x1080 (laptop built-in)_
    * Monitor 2: _1366x768 (external)_
* Python version: _3.6.5_
* MSS version: __3.2.0__

For GNU/Linux users:

* Display server protocol and version, if known: __X server__ _(X.Org version: 1.19.6)_
* Desktop Environment: _Gnome 3_
* Composite Window Manager name and version: __???__ (The default one, I guess?)

### Description of the warning/error
This error occurs when the a monitor that is out of bounds is specified in the `grab()` call.
```python
from mss import mss

screenshotter = mss()
monitor = {
    "left": -30,
    "top": 0,
    "width": 100,
    "height": 100,
}
try:
    screenshotter.grab(monitor)  # X Error
except BaseException as e:
    # Never reached
    print("Handling error")
```
Using a monitor like `{"left": 0, "top": 0, "width": 10000, "height": 100}` has the same result.
### Full message
```stderr
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  73 (X_GetImage)
  Serial number of failed request:  91
  Current serial number in output stream:  91
```
The message is printed to `stderr`, then the Python interpreter quits with exit code `1`.

### Other details
I discovered this because I let my program generate a box based on the location of the cursor. Then I take the screenshot of only the cursor area.

While it might be better for efficiency not to check the bounds of the monitor, I personally don't think causing this type of hard crash is acceptable. It's up to you. If you decide not to check bounds, I think a mention of this in the documentation would be appropriate.

BoboTiG:

- Fixes #61 

It introduces a general error handler on GNU/Linux to prevent any interpreter crash. That's said, all Xlib or Xrand functions will raise a `ScreenShotError` exception like:

    mss.exception.ScreenShotError: XGetImage() failed: BadFont (invalid Font parameter)

`details` will contain all XError details:

```python
{'args': (<mss.linux.LP_Display object at 0x7fe2484cc9d8>,
          <mss.linux.LP_Display object at 0x7fe2484ccae8>,
          -30,
          0,
          100,
          100,
          16777215,
          2),
 'retval': <mss.linux.LP_XImage object at 0x7fe2484ccb70>,
 'xerror': 'BadFont (invalid Font parameter)',
 'xerror_details': {'error_code': 7,
                    'minor_code': 0,
                    'request_code': 0,
                    'serial': 422,
                    'type': 0}}
``` 

- [x] Tests added/updated
- [x] Documentation updated

Is your code right?

- [x] PEP8 compliant
- [x] `flake8` passed (bonus)


RedFantom:
This looks good, it works. You published this as I was working on a [fix for myself](https://github.com/RedFantom/python-mss/commit/7e26de184b1e6a0800231b01451f794087a76f73), but this prevents an interpreter crash for all possible errors that can occur with `XGetImage`, which is a lot better.

Given that your last commit contains a code clean-up, I suppose I should mention this: You've got a pair of redundant parentheses in line 392 of `linux.py`.

As for the error messages, I suppose you could make a `Dict[int, str]` that contains some more descriptive error messages for each error code and pass that message as an additional argument to the error message raised.

Thank you for your rapid fix :+1: .

BoboTiG:
The patch is quite good because it will catch all errors of of xlob/xrand functions, not only XGetImage ;)

BoboTiG:
I updated the patch: removed the clean-up commit (will be done after, in a separate PR), and added a textual error message. Read the updated description again and let me know your thoughts.

RedFantom:
That looks great! I'd say this is verbose enough to be able to easily figure out the cause of the error.

PR #103, opened 2019-02-07
MSS: improve type annotations and add CI check

### LINKED ISSUE #101: Use official type annotations
Type annotations are stored in comments, but now that we dropped the Python 2.7 support, we can use them officially. It will fix flake8 issues on the CI in the same time.

We also should add MyPy to the CI in its own Tox testenv. It will be run after the lint one.

BoboTiG:

- Improve type annotations
- Add CI check

- [x] Documentation updated

Is your code right?

- [x] PEP8 compliant
- [x] `flake8` passed


BoboTiG:
Retry Deepcode

PR #117, opened 2019-03-13
Windows DPI bugfix

SergeyKalutsky:

- Fixes #
- Changed SetProcessDpiAwareness attribute call on window 8+ to correct one
- Changed logic to if statements instead of wrapping everything in exeptions
- Put code in a separate method instead of calling from __init__ method.(I really spent a long time trying
to understand what's wrong because the code which is related to the problem has no logical relationships with monitor method. If not fot the comments I wouldn't find the bug)

### DIFF
--- a/mss/windows.py
+++ b/mss/windows.py
@@ -198,9 +198,11 @@ class MSS(MSSMixin):
         """
 
         release = int(platform.win32_ver()[0])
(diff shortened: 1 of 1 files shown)


BoboTiG:
I will have a look to help on this one, much better I think. Great job :+1:

PR #118, opened 2019-03-13
quick windows dpi bug fix

### LINKED ISSUE #116: windows 10 DPI bug
General information:

* OS name: Windows
* OS version:  10
* OS architecture: 64 bits
* Resolutions:
    * Monitor 1: 1366:720
    * Monitor 2: 1920x1080
* Python version: 3.6
* MSS version: 4.0.2

Hi, I've noticed incorrect monitor capture with mss on windows 10 with multiple monitors.
The problem I believe in windows.py:
https://github.com/BoboTiG/python-mss/blob/9b23ed594d9b5deb35ba05bb21e2cb04d0cdc73f/mss/windows.py#L101-L111

ctypes.WinDLL("user32") doesn't have SetProcessDpiAwareness method on windows 10, it causes an AttributeError. The correct code would be ctypes.windll.shcore.SetProcessDpiAwareness(2), to scale each monitor:
 
```
PROCESS_PER_MONITOR_DPI_AWARE = 2
    /*  Per monitor DPI aware. This app checks for the DPI when it is
        created and adjusts the scale factor whenever the DPI changes.
        These applications are not automatically scaled by the system. */
```

So in short. Because SetProcessDpiAwareness isn't correctly called it always causes an attribute error and falls on user32.SetProcessDPIAware() on windows 10, which causes strange behaviour.

SergeyKalutsky:

- Fixes #116 
- changed SetProcessDpiAwareness to correct call on windows 8+ 

Maybe it should be a separate method?

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [ ] PEP8 compliant
- [ ] `flake8` passed

### DIFF
--- a/mss/windows.py
+++ b/mss/windows.py
@@ -190,7 +190,7 @@ class MSS(MSSMixin):
             pass
 
 
-    def _set_dpi_awarness(self):
+    def _set_dpi_awareness(self):
         """
         Set DPI aware to capture full screen on Hi-DPI monitors
         Automatically scale for DPI changes
@@ -210,7 +210,7 @@ class MSS(MSSMixin):
         # type: () -> Monitors
         """ Get positions of monitors (see parent class). """
 
-        self._set_dpi_awarness()
+        self._set_dpi_awareness()
 
         if not self._monitors:
             # All monitors
--- a/mss/windows.py
+++ b/mss/windows.py
@@ -191,11 +191,8 @@ class MSS(MSSMixin):
             pass
 
     def _set_dpi_awareness(self):
-        """
-        Set DPI aware to capture full screen on Hi-DPI monitors
-        Automatically scale for DPI changes
-        """
-        
+        """ Set DPI aware to capture full screen on Hi-DPI monitors. """
+
         version = sys.getwindowsversion()[:2]
         if version >= (6, 3):
             # Windows 8.1+
@@ -204,7 +201,7 @@ class MSS(MSSMixin):
             #     created and adjusts the scale factor whenever the DPI changes.
             #     These applications are not automatically scaled by the system.
             ctypes.windll.shcore.SetProcessDpiAwareness(2)
-        elif (6,0) <= version < (6, 3):
+        elif (6, 0) <= version < (6, 3):
             # Windows Vista, 7, 8 and Server 2012
             self.user32.SetProcessDPIAware()
 
--- a/mss/windows.py
+++ b/mss/windows.py
@@ -3,6 +3,7 @@ This is part of the MSS Python's module.
 Source: https://github.com/BoboTiG/python-mss
 """
 
+import sys
 import ctypes
 from ctypes.wintypes import (
     BOOL,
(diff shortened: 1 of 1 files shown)


BoboTiG:
Actually, I was thinking of the exact same changes as your older closed PR :)

SergeyKalutsky:
Yea, I tried, then I remembered how messy is whole windows versions are. win8 on winserver 2012 is called 6+ an so on. Will be good to separate in a differen method, but I don't believe we can escape exception wrappig without adding significant amount of code.
Something like that:
https://stackoverflow.com/questions/19128219/detect-windows-8-1-in-python/22325767#22325767

BoboTiG:
I think [sys.getwindowsversion()](https://docs.python.org/3/library/sys.html#sys.getwindowsversion) is exactly what you want, right? It is defined [here](https://github.com/python/cpython/blob/adfffc7343ce7ebc88ec734a803d3247ba8927fb/Python/sysmodule.c#L1137) and seems to be a good candidate.

SergeyKalutsky:
Right, it should work much better. Why didn't I whink about that?)) Thanks for help.

mss/windows.py
```diff
@@ -199,11 +189,29 @@ def close(self):
...
             pass
 
+
+    def _set_dpi_awarness(self):
+        """
+        Set DPI aware to capture full screen on Hi-DPI monitors
+        Automatically scale for DPI changes
+        """
+
+        major, minor = sys.getwindowsversion()[:2]
```
BoboTiG: May I propose something a little bit different? Like:
```python
        version = sys.getwindowsversion()[:2]
        if version >= (6, 3):
            # Windows 8.1+
            # Here 2 = PROCESS_PER_MONITOR_DPI_AWARE, which means:
            #     per monitor DPI aware. This app checks for the DPI when it is
            #     created and adjusts the scale factor whenever the DPI changes.
            #     These applications are not automatically scaled by the system.
            ctypes.windll.shcore.SetProcessDpiAwareness(2)
        elif (6,0) <= version < (6, 3):
            # Windows Vista, 7, 8 and Server 2012
            self.user32.SetProcessDPIAware()
```

BoboTiG:
You can add youself to the `CONTRIBUTORS` file if you want ;)

mss/windows.py
SergeyKalutsky: well, that much better))

SergeyKalutsky:
No, thanks)) 
It's first time I contribute to a major python library, experience is a good reward in and of itself

mss/windows.py
```diff
@@ -201,11 +189,31 @@ def close(self):
...
+        elif (6,0) <= version < (6, 3):
+            # Windows Vista, 7, 8 and Server 2012
+            self.user32.SetProcessDPIAware()
+
     @property
     def monitors(self):
         # type: () -> Monitors
         """ Get positions of monitors (see parent class). """
 
+        self._set_dpi_awareness()
```
BoboTiG: Shouldn't we move this call inside `__init__()`?

SergeyKalutsky: ok

BoboTiG:
Thank you very much @SergeyKalutsky !

PR #203, opened 2021-07-28
setup: fix three setuptools warnings

ArchangeGabriel:
UserWarning: Usage of dash-separated 'author-email' will not be supported in future versions. Please use the underscore name 'author_email' instead
UserWarning: Usage of dash-separated 'home-page' will not be supported in future versions. Please use the underscore name 'home_page' instead
UserWarning: Usage of dash-separated 'zip-safe' will not be supported in future versions. Please use the underscore name 'zip_safe' instead

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,11 +2,11 @@
 name = mss
 version = 6.1.0
 author = Mickaël 'Tiger-222' Schoentgen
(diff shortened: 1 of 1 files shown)


BoboTiG:
Thanks @ArchangeGabriel  :)

Issue #213, opened 2022-08-06
Pytest --strict option is deprecated


CTPaHHuK-HEbA:
Run pytest

====== warnings summary =============================================================================================== C:\Program Files\Python310\lib\site-packages\_pytest\config\__init__.py:1198                                                                                                                                      C:\Program Files\Python310\lib\site-packages\_pytest\config\__init__.py:1198: PytestRemovedIn8Warning: The --strict option is deprecated, use --strict-markers instead.  

Need
```
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ max-line-length = 120
 [tool:pytest]
 addopts =
     --showlocals
-    --strict
+    --strict-markers
```

BoboTiG:
Would you mind opening a PR?

I definitely need to update the CI to use GitHub workflows. Will try to do it in a decent delay.

PR #214, opened 2022-08-08
Fix Pytest deprecated

CTPaHHuK-HEbA:
Pytest --strict option is deprecated #213


BoboTiG:
Thanks @CTPaHHuK-HEbA !

PR #240, opened 2023-04-10
Include documentation sources in sdist

mgorny:

Add documentation sources to the list of files included in sdist archives, in order to make it possible to build offline documentation using them.  This is necessary in order to make it possible for Linux distributions such as Gentoo to be able to use sdists.

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [x] PEP8 compliant
- [x] `flake8` passed

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
-# Include tests files and data
+# Include test files, documentation sources and data
 include mss/tests/*.py
+recursive-include docs/source *
 recursive-include mss/tests/res *


mgorny:
That said, I've noticed that the sdist on PyPI currently has `venv` directory — I don't know how it was added there, so I can't test a "fix" but I suspect you may want to exclude it explicitly.

BoboTiG:
Can you fix the venv stuff at the same time? 🙏🏻

BoboTiG:
> That said, I've noticed that the sdist on PyPI currently has `venv` directory — I don't know how it was added there, so I can't test a "fix" but I suspect you may want to exclude it explicitly.

If you can too, enhance the test-setup.py file to check the outout with wanted, and unwanted, files to prevent regressions. I will do it after your PR if you can't, no pression :)

mgorny:
> Can you fix the venv stuff at the same time? 🙏🏻 I guess it was added in [426826b](https://github.com/BoboTiG/python-mss/commit/426826b02634ca2decd35f931e2299397dcd2dcd), maybe do you know a better fix for such a warning?

I think that's the correct fix but tbh I'm not an expert on setuptools anymore.

> FTR, my setup is:
> 
> ```shell
> python -m venv venv
> . venv/bin/activate
> python -m pip install -r dev-requirements.txt
> python -m build
> ```

This doesn't seem to include `venv` for me either. I suspect the right entry would be either `prune venv` or `recursive-exclude venv *` but I don't want to add it blind if I can't reproduce.

> If you can too, enhance the test-setup.py file to check the outout with wanted, and unwanted, files to prevent regressions. I will do it after your PR if you can't, no pression :)

I'm afraid that's too much work for me right now.

BoboTiG:
I'll merge your PR, and fix the venv stuff right after. Thanks :)

mgorny:
Thank you!

PR #249, opened 2023-04-15
tests: Use PyVirtualDisplay instead of xvfbwrapper

mgorny:

Use the more modern PyVirtualDisplay package instead of xvfbwrapper to run Xvfb.  Most importantly, it uses the more robust approach of starting Xvfb with `-displayfd` and it is actively maintained upstream.  Unfortunately, this does not seem sufficient to entirely eliminate random test failures.

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [x] PEP8 compliant
- [x] `flake8` passed

### DIFF
--- a/src/tests/test_gnu_linux.py
+++ b/src/tests/test_gnu_linux.py
@@ -12,7 +12,7 @@ import mss.linux
 from mss.base import MSSBase
 from mss.exception import ScreenShotError
 
-xvfbwrapper = pytest.importorskip("xvfbwrapper")
+pyvirtualdisplay = pytest.importorskip("pyvirtualdisplay")
 
 PYPY = platform.python_implementation() == "PyPy"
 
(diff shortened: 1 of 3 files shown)


mgorny:
Well, let's see how it fares on CI. Sometimes I still get a single test failure, at other times all pass. Perhaps the only way to really avoid that is to use something like `pytest-rerunfailures` to retry 2 or 3 times on failure.

src/tests/test_gnu_linux.py
```diff
@@ -23,12 +23,8 @@
 
 @pytest.fixture
 def display() -> str:
-    vdisplay = xvfbwrapper.Xvfb(width=WIDTH, height=HEIGHT, colordepth=DEPTH)
-    vdisplay.start()
-    try:
-        yield f":{vdisplay.new_display}"
-    finally:
-        vdisplay.stop()
+    with pyvirtualdisplay.Display(size=(WIDTH, HEIGHT), color_depth=DEPTH) as vdisplay:
```
mgorny, line 26: I've used the context manager approach as that's the simplest way of ensuring that `start()` and `stop()` are called automatically.

mgorny:
Thanks a lot!

BoboTiG:
Thanks to you :)

PR #272, opened 2024-01-25
Windows: capture cursor

toxicrecker:

- Capture cursor implemented for windows

- [x] Tests added/updated
- [x] Documentation updated

Is your code right?

- [x] PEP8 compliant
- [x] `flake8` passed

### DIFF
--- a/src/mss/windows.py
+++ b/src/mss/windows.py
@@ -254,13 +254,13 @@ class MSS(MSSBase):
 
         return self.cls_image(bytearray(self._handles.data), monitor)
 
-    def _cursor_impl(self) -> Optional[ScreenShot]: # works for non monochrome cursors
+    def _cursor_impl(self) -> Optional[ScreenShot]:
         """Retrieve all cursor data. Pixels have to be RGB."""
         _, hcursor, pos_win = win32gui.GetCursorInfo()
         
         hdc = win32ui.CreateDCFromHandle(win32gui.GetDC(0))
         hbmp = win32ui.CreateBitmap()
-        hbmp.CreateCompatibleBitmap(hdc, 36, 36)
+        hbmp.CreateCompatibleBitmap(hdc, 32, 32)
         hdc = hdc.CreateCompatibleDC()
         hdc.SelectObject(hbmp)
         hdc.DrawIcon((0,0), hcursor)
@@ -268,14 +268,16 @@ class MSS(MSSBase):
         bmpinfo = hbmp.GetInfo()
         bmpstr = hbmp.GetBitmapBits(True)
         
-        hotspot = win32gui.GetIconInfo(hcursor)[1:3]
+        _, hotspotx, hotspoty, hbmpmask, _ = win32gui.GetIconInfo(hcursor)
         ratio = ctypes.windll.shcore.GetScaleFactorForDevice(0) / 100
         region = {
-            "left": round(pos_win[0]*ratio - hotspot[0]),
-            "top": round(pos_win[1]*ratio - hotspot[1]),
+            "left": round(pos_win[0]*ratio - hotspotx),
+            "top": round(pos_win[1]*ratio - hotspoty),
             "width": bmpinfo["bmWidth"],
             "height": bmpinfo["bmHeight"]
         }
+        mask_info = win32ui.CreateBitmapFromHandle(hbmpmask).GetInfo()
+        is_monochrome = mask_info["bmHeight"] == mask_info["bmWidth"] * 2
 
         win32gui.DestroyIcon(hcursor)    
         win32gui.DeleteObject(hbmp.GetHandle())
@@ -286,5 +288,13 @@ class MSS(MSSBase):
         data[::4] = raw[2::4]
         data[1::4] = raw[1::4]
         data[2::4] = raw[::4]
-        data[3::4] = raw[3::4]
+        if is_monochrome:
+            for i in range(3, len(data), 4):
+                if data[i-3:i] == b"\x00\x00\x00":
+                    data[i] = 0
+                else:
+                    data[i] = 255
+        else:
+            data[3::4] = raw[3::4]
+
         return self.cls_image(data, region)
--- a/src/mss/windows.py
+++ b/src/mss/windows.py
@@ -155,12 +155,12 @@ class MSS(MSSBase):
         bmi.bmiHeader.biClrImportant = 0  # See grab.__doc__ [3]
         self._handles.bmi = bmi
 
-        ci = CURSORINFO()
-        ci.cbSize = ctypes.sizeof(CURSORINFO)
-        self._handles.ci = ci
+        cursor_info = CURSORINFO()
+        cursor_info.cbSize = ctypes.sizeof(CURSORINFO)
+        self._handles.cursor_info = cursor_info
 
-        iconinfo = ICONINFO()  # 'ii' felt uncomfortable
-        self._handles.iconinfo = iconinfo
+        icon_info = ICONINFO()  # 'ii' felt uncomfortable
+        self._handles.icon_info = icon_info
 
     def close(self) -> None:
         # Clean-up
@@ -294,7 +294,7 @@ class MSS(MSSBase):
 
         return self.cls_image(bytearray(self._handles.data), monitor)
 
-    def _cursor_impl(self) -> ScreenShot:
+    def _cursor_impl(self) -> Optional[ScreenShot]:
         """Retrieve all cursor data. Pixels have to be RGB.
 
         [1] user32.DrawIcon(HDC(memdc), 0, 0, hcursor)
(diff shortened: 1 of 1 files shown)


src/mss/windows.py
```diff
@@ -121,6 +155,13 @@ def __init__(self, /, **kwargs: Any) -> None:
         bmi.bmiHeader.biClrImportant = 0  # See grab.__doc__ [3]
         self._handles.bmi = bmi
 
+        ci = CURSORINFO()
```
BoboTiG: Could you use more explicit name? `cusor_info` seems a good one.

src/mss/windows.py
```diff
@@ -121,6 +155,13 @@ def __init__(self, /, **kwargs: Any) -> None:
...
+        ci.cbSize = ctypes.sizeof(CURSORINFO)
+        self._handles.ci = ci
+
+        iconinfo = ICONINFO()  # 'ii' felt uncomfortable
```
BoboTiG:
```suggestion
        icon_info = ICONINFO()
```

src/mss/windows.py
```diff
@@ -200,7 +241,7 @@ def _callback(monitor: int, data: HDC, rect: LPRECT, dc_: LPARAM) -> int:
         callback = MONITORNUMPROC(_callback)
         user32.EnumDisplayMonitors(0, 0, callback, 0)
 
-    def _grab_impl(self, monitor: Monitor, /) -> ScreenShot:
+    def _grab_impl(self, monitor: Monitor, /) -> Optional[ScreenShot]:
```
BoboTiG, line 244: I would like to not change that part. If the process fails, then an exception will be raised.

src/mss/windows.py
```diff
@@ -253,6 +294,67 @@ def _grab_impl(self, monitor: Monitor, /) -> ScreenShot:
...
+        (https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-iconinfo)
+        But I cannot find the correct way of getting a bitmap's dimensions, therefore I
+        just eneded up checking if the color bitmap is null.
+
+        [4]
+        The data received using DrawIcon is in the format BGRA but in case of monochrome
+        cursors the alpha value of every pixel is 0 for some reason. Therefore, the alpha
+        value of every non black pixel has to be manually set to 255.
+        """
+        srcdc, memdc = self._handles.srcdc, self._handles.memdc
```
BoboTiG, line 322: I love the documentation 💪🏻 

For the implementation, do we need to copy all from the grab() method? Can it be simplified to only the cursor itself?

BoboTiG (commented):
Great PR, thank you !

Would you mind adding tests like I did in https://github.com/BoboTiG/python-mss/blob/e66199adcdad737ef4d66183083bd26c827090da/src/tests/test_gnu_linux.py 🙏🏻 ?

src/mss/windows.py
toxicrecker: sure, will do

src/mss/windows.py
toxicrecker: yea i will fix that too

src/mss/windows.py
toxicrecker: it was changed for the linux implementation so i did it but i can see why you would like to avoid that

src/mss/windows.py
toxicrecker: i did not care to look into it but now that you said it i will see if something can be done about simplifying it to just the cursor
i will update you once im done

toxicrecker: UPDATE: im afraid it would not be possible since the cursor is a part of the screen's device context, it should not be a huge issue as im making a small 32x32 bitmap and capturing only the cursor's image
im sorry if im wrong, im not an expert at windows api and i would love to learn something new

toxicrecker:
> Great PR, thank you !
> 
> Would you mind adding tests like I did in https://github.com/BoboTiG/python-mss/blob/e66199adcdad737ef4d66183083bd26c827090da/src/tests/test_gnu_linux.py 🙏🏻 ?

I have absolutely no experience in writing tests, sorry
Best I could do is copy paste what you did with linux tests and remove the `hasattr(sct, "xfixes")` assertation, I will do that if you want

toxicrecker:
I was trying to make sense of how your linux tests worked, and I assume that you try taking screenshots of a black screen one with cursor and one without cursor and then check if you find an RGB value other than 0, am I right? And if I'm right how do you run this test? I can't make sense of how you are actually going to run this test in a way it does it's job because the screen is mostly never pitch black, do you have a second completely blank monitor that you use for this test? Sorry if I'm just being dumb I don't have any experience with testing.

src/mss/windows.py
BoboTiG: We are all good on Linux side: https://github.com/BoboTiG/python-mss/blob/v9.0.1/src/mss/linux.py#L424

BoboTiG:
Windows tests will be more tricky, for sure.
I will be able to help on that side 👍🏻

toxicrecker:
> Windows tests will be more tricky, for sure. I will be able to help on that side 👍🏻

Thank you, is there anything else I could help with?

BoboTiG:
> Thank you, is there anything else I could help with?

It's all good for now. I'll check the code, test it o a VM, and get back to the PR when I'll have some time.

src/mss/windows.py
```diff
@@ -254,5 +295,66 @@ def _grab_impl(self, monitor: Monitor, /) -> ScreenShot:
...
+        bits = gdi.GetDIBits(memdc, self._handles.bmp, 0, height, self._handles.data, self._handles.bmi, DIB_RGB_COLORS)
+        if bits != height:
+            raise ScreenShotError("gdi32.GetDIBits() failed.")
+
+        user32.GetIconInfo(hcursor, self._handles.icon_info)  # [2]
+        is_monochrome = self._handles.icon_info.hbmColor is None  # [3]
+        ratio = ctypes.windll.shcore.GetScaleFactorForDevice(0) / 100
+        region = {
+            "left": round(pos_screen.x * ratio - self._handles.icon_info.xHotspot),
+            "top": round(pos_screen.y * ratio - self._handles.icon_info.yHotspot),
```
Pawloland, line 349: Hi, I used this PR in my project and from my testing multiplying by ratio actually messes things up. When i removed it everything worked fine even on strange DPI scalings in multiple different positions of multi monitor setup. For testing I used https://github.com/pavlobu/deskreen with https://www.amyuni.com/forum/viewtopic.php?t=3030

MohammedAlkindi:
I ran this branch on a physical Windows 11 Home machine (Python 3.13), since the thread stalled in 2024 on needing a Windows test. Two results, one structural and one behavioral.

**1. The branch no longer applies to `main`.** `src/mss/windows.py` was deleted in the strategy refactor (#494/#503), so merging gives a modify/delete conflict, and at current `main` the Windows backend's `cursor()` (`src/mss/windows/gdi.py:417`) still just returns `None`:

```
$ git fetch origin pull/272/head:pr272
$ git merge --no-commit --no-ff pr272
CONFLICT (modify/delete): src/mss/windows.py deleted in HEAD and modified in pr272.
```

So this needs a port to the new backend rather than a rebase.

**2. On the branch itself (checked out at 19cb97f, `pip install -e .`), the failure path is silent and composites a wrong image.** My test session happens to be spawned by a scheduler (non-interactive window station), where `GetCursorInfo` legitimately fails — and none of the API calls in `_cursor_impl` check their return values:

```
GetCursorInfo ok=0 flags=0 (1=CURSOR_SHOWING) pos=(0,0) hCursor=None
_cursor_impl -> ScreenShot size=Size(width=32, height=32) pos=(0,0)
non-zero rgb bytes in cursor image: 0 of 3072
```

With `hCursor=NULL`, `DrawIcon` fails silently, `GetIconInfo` leaves the zero-initialized `ICONINFO` untouched, `hbmColor is None` sends it down the monochrome path, and every all-black pixel gets alpha 0 — so in my run `grab()` under `with_cursor=True` came out pixel-identical to a plain grab (0 differing bytes over a 1920×1200 monitor). That looks benign only by accident: instead of returning `None`, the code hands back a phantom 32×32 cursor at (0,0), and the same unchecked path would also draw a cursor the user has hidden (`flags & CURSOR_SHOWING` is never consulted). Two smaller points from reading the same function: per `GetIconInfo`'s documentation the caller must delete the returned `hbmMask`/`hbmColor` bitmaps, which this never does — two GDI objects leaked per grab, and at capture rates that walks into the default 10,000 GDI-object cap in well under two minutes; and the 32×32 size is hardcoded, which breaks with accessibility large-cursor settings.

**What I could not verify:** the happy path with a visible cursor and correct hotspot placement — this session has no interactive cursor state, so a positive-path screenshot test was not possible from here. The findings above don't depend on it.

The mechanism itself (`CURSORINFO` + `DrawIcon`) still looks like the right Windows approach; porting it into `src/mss/windows/gdi.py` with return-value checks on `GetCursorInfo`/`GetIconInfo`, a `CURSOR_SHOWING` gate, and bitmap cleanup seems like the path if there's still interest.

PR #274, opened 2024-02-08
Pixels model correction

Andon-Li:
ScreenShot.pixels returns a list where each element is a tuple representing a row of pixels. Each of these tuples contain "Pixel"'s. The current model incorrectly hints the return value of ScreenShot.pixels as a list containing "Pixel"s.

- Model for Pixels changed to reflect the true nature of the data

- [X] Tests added/updated
- [X] Documentation updated

Is your code right?

- [X] PEP8 compliant
- [X] `flake8` passed

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -312,9 +312,9 @@ Properties
 
     .. attribute:: pixels
 
-        List of RGB tuples.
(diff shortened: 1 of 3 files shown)


BoboTiG:
Thank you @Andon-Li 🍾

PR #300, opened 2024-11-01
feat: Python 3.14 support

BoboTiG:


BoboTiG:
Requires Windows wheels for Pillow.

PR #417, opened 2025-10-30
Check the server for Xrandr support version

jholveck:
We currently only check the client libXrandr to see if it supports the XRRGetScreenResourcesCurrent function.  To make sure the function will work, we also need to check the server's Xrandr version.

That said, Xrandr 1.3 was released in 2009, and the code that checks for its existence (849e1fe) was added in 2019.  Today, the version check might no longer be necessary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -182,7 +182,8 @@ ignore = [
   "FBT001",   # boolean-typed positional argument in function definition
   "PLR2004",  # magic value used in comparison
   "S101",     # use of `assert` detected
-  "S602",     # `subprocess` call with `shell=True`
-  "S603",     # `subprocess` call
+  #"S602",     # `subprocess` call with `shell=True`
+  "S603",     # `subprocess` call: check for execution of untrusted input
+  "S607",     # `subprocess` call without explicit paths
   "SLF001",   # private member accessed
 ]
--- a/src/tests/test_leaks.py
+++ b/src/tests/test_leaks.py
@@ -20,9 +20,9 @@ def get_opened_socket() -> int:
     """GNU/Linux: a way to get the opened sockets count.
     It will be used to check X server connections are well closed.
     """
-    cmd = f"lsof -U | grep {PID}"
-    output = subprocess.check_output(cmd, shell=True)
-    return len(output.splitlines())
+    output = subprocess.check_output(["lsof", "-a", "-U", "-Ff", f"-p{PID}"])
+    # The first line will be "p{PID}".  The remaining lines start with "f", one per open socket.
+    return len([line for line in output.splitlines() if line.startswith(b"f")])
 
 
 def get_handles() -> int:


BoboTiG:
Thanks @jholveck!

About the failing CI, only Python 3.14 on Windows is known to never pass yet: others are passing normally.

BoboTiG:
Overall seems good to me, would you mind fixing tests + adding line in the changelog? After that, I will merge :)

jholveck:
The CI bot seems to indicate that the Linux tests are failing too.  Is that not what you're seeing?

I'll take another look at this change and update the PR.

BoboTiG:
Yup, I see failling Linux tests. I assumed you'll see them too and fix them :)

jholveck:
Ah, I misunderstood your earlier comment.  Yes, of course I'll fix these.

How do you usually run tests from the command line before your commit?  With `xvfb-run python -m pytest`?

BoboTiG:
This is specific to the CI where a X server is not necessarily available. On your machine, `python -m pytest` should be enough (https://python-mss.readthedocs.io/developers.html#testing for full instructions).

jholveck:
Ok!  After looking at this, I think that the test itself is faulty.

As written, it will use `del sct.xrandr.XRRGetScreenResourcesCurrent` and then make sure that a `grab()` can succeed.  However, `del` seems to not work as expected to delete a function from a ctypes library, so that test is still following the fast path.  (You can test this easily by putting `assert not hasattr(sct.xrandr, "XRRGetScreenResourcesCurrent")` right after the `del`.)

I'll come up with a better way to structure this test.

jholveck:
I've verified that the test itself is at issue, but the specific path through which it was previously passing is amusing.

I've got a patch with the details, and a much more stringent set of tests, but it's late in California and I want to review it with fresh eyes before I commit.

BoboTiG:
Next round tomorrow :)

jholveck:
I've updated the test.  You might find the description in the commit log interesting.

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
+        slow_spy = spy_and_patch(monkeypatch, sct.xrandr, "XRRGetScreenResources")
+        # If we just delete sct.xrandr.XRRGetScreenResourcesCurrent, it will get recreated automatically by ctypes
+        # the next time it's accessed.  A Mock will remember that the attribute was explicitly deleted and hide it.
+        mock_xrandr = NonCallableMock(wraps=sct.xrandr)
+        del mock_xrandr.XRRGetScreenResourcesCurrent
+        monkeypatch.setattr(sct, "xrandr", mock_xrandr)
+        assert not hasattr(sct.xrandr, "XRRGetScreenResourcesCurrent")
         screenshot_with_slow_fn = sct.grab(sct.monitors[1])
 
+    fast_spy.assert_not_called()  # fast path used
```
BoboTiG: The comment is not aligned with the code. I think you can remove those comments completely since `.assert_*()` are explicit enough.

BoboTiG:
Great tests, and thank you for the explanation. It's a lesson about properly testing things 😅

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
-def test_missing_fast_function_for_monitor_details_retrieval(display: str) -> None:
+def test_fast_function_for_monitor_details_retrieval(display: str, monkeypatch: pytest.MonkeyPatch) -> None:
     with mss.mss(display=display) as sct:
         assert isinstance(sct, mss.linux.MSS)  # For Mypy
         assert hasattr(sct.xrandr, "XRRGetScreenResourcesCurrent")
+        fast_spy = spy_and_patch(monkeypatch, sct.xrandr, "XRRGetScreenResourcesCurrent")
+        slow_spy = spy_and_patch(monkeypatch, sct.xrandr, "XRRGetScreenResources")
         screenshot_with_fast_fn = sct.grab(sct.monitors[1])
 
+    fast_spy.assert_called()  # fast path used
```
BoboTiG:
```suggestion
    fast_spy.assert_called()
```

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
+    slow_spy.assert_not_called()  # slow path not used
```
BoboTiG:
```suggestion
    slow_spy.assert_not_called()
```

src/tests/test_gnu_linux.py
BoboTiG:
```suggestion
    fast_spy.assert_not_called()
```

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
+        # If we just delete sct.xrandr.XRRGetScreenResourcesCurrent, it will get recreated automatically by ctypes
+        # the next time it's accessed.  A Mock will remember that the attribute was explicitly deleted and hide it.
+        mock_xrandr = NonCallableMock(wraps=sct.xrandr)
+        del mock_xrandr.XRRGetScreenResourcesCurrent
+        monkeypatch.setattr(sct, "xrandr", mock_xrandr)
+        assert not hasattr(sct.xrandr, "XRRGetScreenResourcesCurrent")
         screenshot_with_slow_fn = sct.grab(sct.monitors[1])
 
+    fast_spy.assert_not_called()  # fast path used
+    slow_spy.assert_called()  # slow path not used
```
BoboTiG:
```suggestion
    slow_spy.assert_called()
```

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
+        return 1
+
+    with mss.mss(display=display) as sct:
+        assert isinstance(sct, mss.linux.MSS)  # For Mypy
+        monkeypatch.setattr(sct.xrandr, "XRRQueryVersion", fake_xrrqueryversion)
+        fast_spy = spy_and_patch(monkeypatch, sct.xrandr, "XRRGetScreenResourcesCurrent")
+        slow_spy = spy_and_patch(monkeypatch, sct.xrandr, "XRRGetScreenResources")
+        screenshot_with_slow_fn = sct.grab(sct.monitors[1])
+
+    fast_spy.assert_not_called()  # fast path used
```
BoboTiG:
```suggestion
    fast_spy.assert_not_called()
```

src/tests/test_gnu_linux.py
```diff
@@ -133,20 +143,70 @@ def test__is_extension_enabled_unknown_name(display: str) -> None:
...
+    slow_spy.assert_called()  # slow path not used
```
BoboTiG:
```suggestion
    slow_spy.assert_called()
```

PR #418, opened 2025-10-30
X11 typing and error message improvements

jholveck:
Several places, a POINTER(Display) was declared, when the actual object was an XID (opaque identifier).  In one place, a
POINTER(XWindowAttributes) was declared, instead of an XID.  This is also why self._handles.drawable had to be cast.

Also, the XErrorEvent structure had its fields in the wrong order.  This meant that error messages in exception details were incorrect.

I'm keeping self._handles.drawable, even though it's currently the same as self._handles.root, in anticipation that we may later support capturing individual windows.

Since the XErrorEvent structure is now fixed, we can improve the error messages.  Rather than just saying that call such-and-such failed, we can also report in the error string what the actual error message is, and other details.

### DIFF
--- a/src/mss/linux.py
+++ b/src/mss/linux.py
@@ -11,10 +11,10 @@ from ctypes import (
     CFUNCTYPE,
     POINTER,
     Structure,
+    _Pointer,
     byref,
     c_char_p,
     c_int,
-    c_long,
     c_short,
     c_ubyte,
     c_uint,
@@ -39,7 +39,6 @@ if TYPE_CHECKING:  # pragma: nocover
 __all__ = ("MSS",)
 
 
-XID = c_ulong
 X_FIRST_EXTENSION_OPCODE = 128
 PLAINMASK = 0x00FFFFFF
 ZPIXMAP = 2
@@ -47,6 +46,31 @@ BITS_PER_PIXELS_32 = 32
 SUPPORTED_BITS_PER_PIXELS = {
     BITS_PER_PIXELS_32,
 }
+
+
+class XID(c_ulong):
+    """X11 generic resource ID
+    https://tronche.com/gui/x/xlib/introduction/generic.html
+    https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/include/X11/X.h#L66
+    """
+
+
+class XStatus(c_int):
+    """Xlib common return code type
+    This is Status in Xlib, but XStatus here to prevent ambiguity.
+    Zero is an error, non-zero is success.
+    https://tronche.com/gui/x/xlib/introduction/errors.html
+    https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/master/include/X11/Xlib.h#L79
+    """
+
+
+class XBool(c_int):
+    """Xlib boolean type
+    This is Bool in Xlib, but XBool here to prevent ambiguity.
+    0 is False, 1 is True.
+    https://tronche.com/gui/x/xlib/introduction/generic.html
+    https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/master/include/X11/Xlib.h#L78
+    """
 
 
 class Display(Structure):
@@ -212,14 +236,14 @@ class XWindowAttributes(Structure):
         ("backing_store", c_int),  # NotUseful, WhenMapped, Always
         ("backing_planes", c_ulong),  # planes to be preserved if possible
         ("backing_pixel", c_ulong),  # value to be used when restoring planes
-        ("save_under", c_int),  # boolean, should bits under be saved?
+        ("save_under", XBool),  # boolean, should bits under be saved?
         ("colormap", XID),  # color map to be associated with window
-        ("mapinstalled", c_int),  # boolean, is color map currently installed
+        ("mapinstalled", XBool),  # boolean, is color map currently installed
         ("map_state", c_uint),  # IsUnmapped, IsUnviewable, IsViewable
         ("all_event_masks", c_ulong),  # set of events all people have interest in
         ("your_event_mask", c_ulong),  # my event mask
         ("do_not_propagate_mask", c_ulong),  # set of events that should not propagate
-        ("override_redirect", c_int),  # boolean value for override-redirect
+        ("override_redirect", XBool),  # boolean value for override-redirect
         ("screen", POINTER(Screen)),  # back pointer to correct screen
     )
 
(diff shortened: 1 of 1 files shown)


BoboTiG:
Wow, it seems I did it quite wrong for so looooong! Thanks for catching this!

jholveck:
No worries!

It seems that there are two issues with this patch.  First, I was relying on the recent Python f-string semantics, in which you can use the same outer quote character within a {} section of an f-string.

Second, I see that my changes to the Display structure cause one of the tests to fail.  I'll see where the test used to raise the TypeError and whether the test needs to change, or my code.  I'm not quite clear on why that test exists, but I'll see what's up.

I'll update the PR when I've fixed these.

src/mss/linux.py
```diff
@@ -194,6 +202,25 @@ class XWindowAttributes(Structure):
...
+class XError(ScreenShotError):
+    def __str__(self):
+        msg = super().__str__()
+        # We use something similar to the default Xlib error handler's format, since that's quite
+        # well-understood.  The original code is in
+        # https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/master/src/XlibInt.c?ref_type=heads#L1313
+        # but we don't try to implement most of it.
+        msg += (
+            f"\nX Error of failed request:  {self.details["error"]}"
+            f"\n  Major opcode of failed request:  {self.details["request_code"]} ({self.details["request"]})")
```
BoboTiG:
```suggestion
            f"\n  Major opcode of failed request:  {self.details['request_code']} ({self.details['request']})")
```

+ same for other lines should fix the syntax error.

BoboTiG: Oh, I read too fast. You know this ofc.

I didn't know about the recent support of same quotes within f-strings. That will be convenient, but it's too early yet.

jholveck: > Oh, I read too fast. You know this ofc.

Don't give me too much credit!

I previously wrote my f-strings assuming that I had to use different outer than inner quotes, like you obviously have to do with any template mechanism without syntactic language support.  Recently, I happened to have used the same inner quotes as I had for the outer quotes, and saw that the syntax highlighter was cool with it.  When I ran the code, it worked.  I then just figured that f-strings had always been spec'd that way, so I started using whatever quote syntax was natural and uniform.

I'm glad that your checks run against multiple Python versions, to call out issues like this!  New thing learned!

I'll send a fixed version tomorrow.

src/mss/linux.py
```diff
@@ -63,11 +71,11 @@ class XErrorEvent(Structure):
     _fields_ = (
         ("type", c_int),
         ("display", POINTER(Display)),  # Display the event was read from
+        ("resourceid", XID),  # resource ID
         ("serial", c_ulong),  # serial number of failed request
         ("error_code", c_ubyte),  # error code of failed request
         ("request_code", c_ubyte),  # major op-code of failed request
         ("minor_code", c_ubyte),  # minor op-code of failed request
-        ("resourceid", c_void_p),  # resource ID
```
BoboTiG, line 70: https://tronche.com/gui/x/xlib/event-handling/protocol-errors/default-handlers.html and https://www.x.org/releases/X11R7.7/doc/man/man3/XErrorEvent.3.xhtml state that `resourceid` is at the end. Maybe did I misinterpret, or it is too old?

BoboTiG: Well, https://www.mankier.com/3/XErrorEvent, https://github.com/mirror/libX11/blob/ff8706a5eae25b8bafce300527079f68a201d27f/include/X11/Xlib.h#L927, and `/usr/include/X11/Xlib.h` on my Debian system, are aligned with your version.

It seems sources I used at the time were too old :)

jholveck: That's fascinating; terrific find!  I see the same thing in https://www.x.org/releases/current/doc/man/man3/XErrorEvent.3.xhtml as well.

But the source of truth is ultimately going to be /usr/include/X11/Xlib.h, which on my computer puts resourceid right after display in XErrorEvent.  The manual pages are mostly intended to identify the fields in the structure for a human; the header files are what the compiler will care about for layout.

It's also entirely legal for different platforms or different library versions to declare these with different sizes and order.  I've got one machine that's not even byte-oriented (it uses 40-bit words) but speaks X11, and it obviously uses different ABI layouts.  It could be that older versions of Xlib put resourceid at the end, but more recent versions put it where it is now for better alignment (because of having three unsigned chars before it).  The spec fixes the wire protocol, but the ABI to Xlib - a higher-level interface - can be more flexible.

I've got other proposals I'm working on.  One is about using a C extension or similar to make the appropriate bindings in some circumstances; that would pick up the right header at compile-time, and so the C code would naturally populate or read structs in the right order.

Another proposal I've got in the works is to use xcffib, a Python library that directly uses the X11 protocol definitions to build its functions.  It's very much in the same spirit as XCB, and even similar enough to use XCB manual pages to infer the right functions and arguments.  But being a library designed in and for Python, it doesn't come with all these ctypes issues.

The xcffib library, like XCB,  is different enough from Xlib that I didn't want to throw that into the works right off the bat.  I wanted to first talk to you about how you feel about importing other Python libraries, since you currently don't.

But now that we're facing one of the relevant issues, I'll say that I think that might simplify some parts of the code.

BoboTiG: Lets go easy first :)

Fixing this XErrorEvent is good with your patch. No need right now to rely on some other modules. I like the external-modules-free vision, if we can stick with it, that would be awesome.

jholveck:
Regarding the `test_bad_display_structure` failure:

The reason it no longer fails is because the new code doesn't explicitly refer to `Display` anymore in the construction path.  Previously, the test expected a `TypeError` to be raised when `self._handles.drawable` was cast to a `POINTER(Display)`.  Since that cast is no longer appropriate with the struct changes, the `test_bad_display_structure` doesn't see a TypeError, and fails.

The test was added in commit 8e6b7528, but I'm not clear on what its purpose is.  I'm happy to adapt it to the new code, if you can tell me a bit about what it's trying to test.  Alternatively, if it's no longer relevant, I can just delete it.

src/mss/linux.py
jholveck: Out of curiosity, I went back through the history of Xlib.h.  And surprisingly, the structure always had resourceid right after Display.  This is the definition from X11R1 (1987):
```
typedef struct {
	int type;
	Display *display;	/* Display the event was read from */
	XID resourceid;		/* resource id */
	int serial;		/* serial number of failed request */
	char error_code;	/* error code of failed request */
	char request_code;	/* Major op-code of failed request */
	char minor_code;	/* Minor op-code of failed request */
} XErrorEvent;
```

And today's definition:
```
typedef struct {
	int type;
	Display *display;	/* Display the event was read from */
	XID resourceid;		/* resource id */
	unsigned long serial;	/* serial number of failed request */
	unsigned char error_code;	/* error code of failed request */
	unsigned char request_code;	/* Major op-code of failed request */
	unsigned char minor_code;	/* Minor op-code of failed request */
} XErrorEvent;
```

BoboTiG: Yeah, my source of truth since the beginning was the outdated https://tronche.com/gui/x/xlib/event-handling/protocol-errors/default-handlers.html... I used Xlib.h from time to time when I wasn't finding much on internet, but I should have checked more closely.

jholveck: I've got a new set of changes that I just wrote (haven't pushed yet).  I reviewed all the structures against the headers, added links to the X.Org GitLab, and added links to Tronche's book where appropriate for explanations.

BoboTiG:
I do not get why I added `test_bad_display_structure()`. Lets remove it :)

src/mss/linux.py
jholveck: I see that Tronche's book is a conversion of the Xlib Programming Manual to multi-page HTML.  He also admits that it's incomplete: I do see that there are pages missing from TOCs and such.

Perhaps we would want to change the links to Tronche's book to point to the one-page Xlib Programming Manual at https://www.x.org/releases/current/doc/libX11/libX11/libX11.html instead.

PR #421, opened 2025-11-04
Re-enable ruff check S602

jholveck:
I disabled this temporarily while developing something else, but forgot to re-enable it.

Apologies!

- Fixes no issues

- [x] Tests added/updated : nothing to test
- [x] Documentation updated : nothing to document

Is your code right?

- [x] `./check.sh` passed : existing code was fine with the newly-enabled test


jholveck:
The thread failure on Linux 3.14 seems unrelated to my change, for obvious reasons.  This test sometimes fails for me, in which one of the threads gets an error in XOpenDisplay.  I think that there are some remaining issues with threading as it currently stands, although this might also be an issue with Xvfb.

Issue #251, opened 2023-04-15
tests: fix random segfault on GNU/Linux


BoboTiG:
Such a segfault happens quite often, even after:
- #247
- #248
- #249

```
$ xvfb-run python -m pytest
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64/lib
============================= test session starts ==============================
platform linux -- Python 3.12.0a7, pytest-7.3.1, pluggy-1.0.0 -- /opt/hostedtoolcache/Python/3.12.0-alpha.7/x64/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/python-mss/python-mss
configfile: setup.cfg
plugins: cov-4.0.0
collecting ... collected 77 items

src/tests/test_bgra_to_rgb.py::test_bad_length PASSED                    [  1%]
src/tests/test_bgra_to_rgb.py::test_good_types PASSED                    [  2%]
src/tests/test_cls_image.py::test_custom_cls_image PASSED                [  3%]
src/tests/test_find_monitors.py::test_get_monitors PASSED                [  5%]
src/tests/test_find_monitors.py::test_keys_aio PASSED                    [  6%]
src/tests/test_find_monitors.py::test_keys_monitor_1 PASSED              [  7%]
src/tests/test_find_monitors.py::test_dimensions PASSED                  [  9%]
src/tests/test_get_pixels.py::test_grab_monitor PASSED                   [ 10%]
src/tests/test_get_pixels.py::test_grab_part_of_screen PASSED            [ 11%]
src/tests/test_get_pixels.py::test_grab_part_of_screen_rounded PASSED    [ 12%]
src/tests/test_get_pixels.py::test_grab_individual_pixels PASSED         [ 14%]
[...]

Issue #207, opened 2021-10-14
Display Profile Color Correction


Snipeye:
General information:

* OS name: _MacOS _
* OS version: _10.14.6, but I anticipate it applies to newer versions_
* OS architecture: _64 bits_
* Resolutions:
    * Monitor 1: _1920x1200 screen resolution, screen is retina and sampled at 2x_
* Python version: _3.8.8_
* MSS version: __6.1.0__

### Description of the warning/error

When a capture (using .grab) in macOS, the resultant screenshot is modified according to the selected color profile.  A picture of a pure #FF00FF color resulted in RGB values of [252  40 252] - this agrees with the system color profile tool:

<img width="527" alt="image" src="https://user-images.githubusercontent.com/3460093/137251776-f3498427-c2ce-4c1c-bdfd-ce8fcf12b8dc.png">

A workaround is to set my profile to sRGB IEC61966-2.1, but even that results in some "corruption" ([251   0 255]).  Is there a way we can reverse the color-correction?  I think that CGImage has a link to the decode transform used to change color spaces? (https://developer.apple.com/documentation/coregraphics/cgimage/1454575-decode) but I'm not sure if that's what's being used under the hood for mss, I only stumbled on this today when looking for some fast image capture.

### Full message

None

### Other details

None

jholveck:
You're correct that what you're seeing is due to macOS applying the display’s color profile — usually Display P3 on modern Macs. MSS just returns raw, untagged RGB values from the compositor, so without an ICC profile attached, those numbers have no defined color meaning.

That’s not technically “corruption,” just the expected result of capturing pixels that have already been color-managed for your display.

That said, I'm not really sure that reversing the conversion is always the best idea.

The values that MSS returns are just raw, untagged RGB values.  Without a colorspace attached (like "Color LCD", which is more technically known as Display P3), you don't actually have information about what colors these represent.  As you've observed, this can be a problem.

However, converting to an arbitrary colorspace, like sRGB, for MSS's output has its downsides too.  For instance, suppose that you have a program that renders in a wide gamut: you might be using Photoshop, or viewing photos from the iPhone camera, for instance.  Converting this to sRGB would mean that all the out-of-gamut colors that look so vivid on your monitor would be crushed.

[...]

PR #426, opened 2025-11-07
Add an XCB-based backend

### LINKED ISSUE #425: Ease X11 maintenance and thread safety by using XCB or xcffib
Right now, we’ve got a lot of ctypes code to bind against Xlib (`libX11`) and friends.

These days, Xlib [isn’t recommended for new code](https://www.remlab.net/op/xlib.shtml).  It was meant to be programmer-friendly, but also has some limitations.  It hides a lot of the implementation details, making it sometimes unpredictable.  In 2001, the X Windows maintainers created a new library, called [XCB](https://en.wikipedia.org/wiki/XCB).  This is closer to the direct X11 protocol than Xlib is.  Most notably, XCB makes multithreading easier, adds asynchronous support, and is generally easier to reason about and use.

XCB connections can be reused by multiple threads without issues; this is a ground-up feature of the XCB design.  By contrast, Xlib's threading support is tacked on, and very difficult to use correctly.  As an example, we currently don't: we don't call XInitThreads, and while we use separate per-thread connections, we don't do anything to lock Xlib's global state.  This probably isn't causing us any practical problems, but it's still an example of how hard it is to get Xlib's threading right.

Many of Xlib's calls are asynchronous, and errors are often reported by future calls.  (Xlib has a facility to operate synchronously, but this is all-or-nothing, and intended to only be used for debugging.  Enabling this in normal usage is far too slow.)  While most of the existing MSS calls to Xlib are synchronous, some proposals (like #96) do involve asynchronous calls, such as `XShmAttach`.  For such calls, Xlib defaults to asynchronous for most calls with no choice.  Dealing with errors in asynchronous calls is a well-known frustration among Xlib veterans.

XCB provides both synchronous and asynchronous options for all its calls.  A program can choose to check for errors when it makes a call, and the error can be raised at that time.  Alternatively, for better performance, a thread can issue several calls in a row asynchronously, and then read results and check the errors when it needs to.  (The results can even come much later: I've got a proposal in the works for an option to let MSS always have one screenshot request in-flight, only reading the results when the program is ready for them.)  The calls made by one thread don't impact the errors or results seen by other threads, so you don't have to worry about attributing errors to the right threads, or blocking other threads using the same connection.

XCB does have a good Python binding, based on cffi: [xcffib](https://github.com/tych0/xcffib).  Alternatively, ctypes can be used with libxcb.

I propose that MSS switches from using Xlib to XCB.  (I would prefer to use xcffib instead of ctypes with libxcb, but we can discuss that.)  This eases the issues we see with multithreading and error handling, and lets us simplify the code and make it more maintainable.

jholveck:

This adds a ctypes-based module to use XCB.

* The mss.linux.MSS constructor is now a factory, which takes an optional `backend` keyword parameter.
* The previous mss.linux implementation has now been designated as the "xlib" backend.  Per #425, this will have "legacy" status, and new functionality like XShmGetImage won't be implemented there.
* A new XCB-based backend, named "getimage", has been added.  This is functionally the same as the existing "xlib" backend, in almost all respects, but has a different implementation.

Notable improvement: I previously was seeing periodic thread failures in `test_thread_safety`.  In one scenario (Docker container, Python 3.9, Xvfb), I tested 84% failures in `test_thread_safety`.  With the XCB-based backend, I saw no failures.  (See also #251 and the test failure I mentioned in #421.)

The performance is, in my brief testing, the same as the existing xlib implementation.  This PR is about making clearer code and improving thread safety, to pave the way for future improvements.

- Fixes #425

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [ ] `./check.sh` passed

### DIFF
--- a/src/tests/test_setup.py
+++ b/src/tests/test_setup.py
@@ -101,6 +101,7 @@ def test_sdist() -> None:
         f"mss-{__version__}/src/tests/third_party/__init__.py",
         f"mss-{__version__}/src/tests/third_party/test_numpy.py",
         f"mss-{__version__}/src/tests/third_party/test_pil.py",
+        f"mss-{__version__}/src/xcbproto/README.md",
         f"mss-{__version__}/src/xcbproto/gen_xcb_to_py.py",
         f"mss-{__version__}/src/xcbproto/randr.xml",
         f"mss-{__version__}/src/xcbproto/render.xml",


jholveck:
As you can see, this is currently a draft PR.  I'm just putting it out to show the work in progress.

The GNU/Linux-specific tests that could easily be made to work with the new backend have been adjusted, but there's more to be done still.

* The way the libraries are initialized needs to be rethought; I'll want to talk with @BoboTiG about that.
* Many tests in `test_gnu_linux` used monkey-patches that are specific to the Xlib implementation.  I've arranged for those to only use the Xlib implementation; I need to add corresponding tests for the XCB implementation.
* The tests outside of `test_gnu_linux` need to test both backends, as well as the different backends we may add to other OSs.  I need to add the right Pytest code to parameterize these.
* I need to fix the (currently-failing) tests on other OS's, of course.  That will come with how I handle the backend parameterization.
* This code introduces new automatic memory management aspects that need tests assigned to them.
* The bulk of the `xcb.py` code is just hand-translations of the [XCB protocol specification XML files](https://gitlab.freedesktop.org/xorg/proto/xcbproto).  Hand-translation can be error-prone (see #418).  Since the XCB specs are in XML, then the bulk of this code can be auto-generated, like xcffib does.  This can be done manually as needed, before shipping the distribution, so end users won't have to do this.  (See #425 for a discussion of why we're not using xcffib.)
* Even if we don't auto-generate the code, some of the repetitive parts can still be hoisted into higher-order functional programming, to reduce the risk of making errors in copying.
* There are some parts that I worsened, to satisfy ruff.  I want to revisit those.
* The proof-of-concept code in `xcb.py:main` includes some additional functionality should be integrated into the MSS class, either in this PR or in separate branches.  We also should delete that code before merging.
  * Additional testing on the X11 visual, to make sure it's really in the format we expect.  (This is probably the only part we'll merge in with this PR.)
  * Support for capturing an individual window, without needing its location.
  * Identifying whether the alpha channel is meaningful or not.  (This is only relevant with individual-window capture; there aren't many screens that can turn transparent!)

jholveck:
> The way the libraries are initialized needs to be rethought; I'll want to talk with @BoboTiG about that.

So, @BoboTiG, let me get your thoughts on this.

The Xlib-based code will re-load libX11 etc. each time an MSS instance is constructed.  (Under the hood, the library doesn't actually get reloaded; the same library that was previously loaded is reused (see dlopen(3)).  However, a new CDLL instance with all-new _Funcptrs is created, so the .argtypes etc. that you've set are new.)

In my current version of the code, the library is loaded and initialized (i.e., argtypes etc. are set) only once, and that is reused by every future MSS instance.

Do you have a strong preference for one or the other?

BoboTiG:
Wow, that's quite something you've done here! That's really good work! 🥂 

About ruff, it is here to help us. When it is more a pain than it should be, you can workaround the rules either by adding specific skips for `xcb.py` like done in https://github.com/BoboTiG/python-mss/blob/7cb3dcc70677458a039e5b0d1e0430ae9a0c10fc/pyproject.toml#L181, or skip specific rules in the code directly using `noqa: xxx`.
I do not like such a tool to force us changing our core implementation. If you have better code, then use it and tell ruff to ignore it.

About the backend name, I would advocate for xlib and xcb. We can support getimage as it is in the code right now, but for end users I would not promote it. It's like adding xgetimage as an alias to xlib.

About testing, maybe could we add parametrized tests "by default"? I mean, maybe pytest allowes us to create a factory to inject parameters, so that we would only need to set backends to test in one place, then the tests runner will deduplicate everything automatically for all backends? Not sure we want that, just asking.

I have not much more to say about the PR, this is very interesting, and I'm eager to see where it will go at the end on the performances level + thread-safety levels.

BoboTiG:
> In my current version of the code, the library is loaded and initialized (i.e., argtypes etc. are set) only once, and that is reused by every future MSS instance.
> 
> Do you have a strong preference for one or the other?

Not really. The later *seems* better.
If this does not introduce more tricky errors with the errors handler then go with your preference.

jholveck:
I've added some test parameterization.

I now have a `backend` fixture, and an `mss_impl` fixture.  The `backend` fixture simply iterates through the backends available for that OS.  The `mss_impl` fixture will give you a callable that will, effectively, run `mss(*args, display=os.environ["DISPLAY"], backend=backend, **kwargs)`, where it's iterating through the backends.

In other words, you can replace this:
```
    with mss.mss(display=os.environ["DISPLAY"]):
        ....
```
with this:
```
    with mss_impl():
        ....
```

For the cases that need more fine-grained work, then I've used the `backend` fixture directly instead.

I see that some tests use `display=os.environ["DISPLAY"]`, some don't give a `display` argument, and some use pyvirtualdisplay to set up an Xvfb and use that.  Is there a particular reason for the differences, or is that a historical accident?

I can change `mss_impl` to use pyvirtualdisplay if it's available, and let the program use the default display (which should be `os.environ["DISPLAY"]` on Linux) if pyvirtualdisplay isn't installed.  I'm not really sure if there's a reason for using `display=os.environ["DISPLAY"]` in any case, except a test to specifically see that the `display=` parameter is properly honored if it's correct.

src/mss/linux/xgetimage.py
```diff
@@ -181,10 +184,7 @@ def _cursor_impl_check_xfixes(self) -> bool:
         # We can work with 2.0 and later, but not sure about the
         # actual minimum version we can use.  That's ok; everything
         # these days is much more modern.
-        if (reply.major_version, reply.minor_version) < (2, 0):
-            return False
-
-        return True
+        return not (reply.major_version, reply.minor_version) < (2, 0)
```
BoboTiG: Why not removing the negation?

```suggestion
        return (reply.major_version, reply.minor_version) >= (2, 0)
```

jholveck: Heh.  That was a Ruff auto-fix.  I'll include your suggestion in my next batch.

BoboTiG:
> Is there a particular reason for the differences, or is that a historical accident?

Right now, I can't tell. I can be to workaround CI issues, as well as testing different `DISPLAY` envars found in user environments.

BoboTiG:
About backend names, I was more proposing to have:
- "xlib" (+ "default" now, which makes sense)
- "xcb"

And get rid of xgetimage and getimage since that's not really backend names, it's too specific.

WDYT?

jholveck:
> About the backend name, I would advocate for xlib and xcb. We can support getimage as it is in the code right now, but for end users I would not promote it. It's like adding xgetimage as an alias to xlib.

Here's my thought about the backend name.

Here's the backends I anticipate having:
* Xlib, with GetImage (existing code)
* XCB, with GetImage (this code)
* XCB, with ShmGetImage
* XCB, with XComposite

As an aside, the name of the X11 protocol message is GetImage, the name of the Xlib function is XGetImage, and the name of the XCB function is xcb_get_image.  In all these cases, though, it's often just called "XGetImage" in conversation.

The reason I envision three different XCB-based backends is because they apply under different circumstances.

* GetImage is the slowest, but it's the only one that always is applicable.  In particular, it's the only one that works with remote connections, such as over ssh.
* XComposite is the fastest by far, but only can capture single windows, and pretty much only makes sense for capturing to a GPU buffer.
* ShmGetImage is the best for full-screen capture or capturing to a CPU buffer on a local machine, a very common use case.

I figure the factory can dispatch to the correct backend automatically, based on what the user has requested.

Now, with these four backends, the naming that made the most sense to me was to name the XCB-based backends (which will be the preferred backends) based on the mechanism they're using.  Since we're also keeping an Xlib-based backend, but not adding new mechanisms to that, then it seemed like "xlib" was the most sensible name for it.

Does that make sense to you?

BoboTiG:
Ah, you're right! That makes more sense to me now.

jholveck:
Glad to hear it!  Sorry I wasn't more clear about my reasoning before.  I've waffled a lot about the backend naming, and I'm happy to change things up before we merge.  That's the point of this being a draft PR!

One downside here is that the user will need to choose these options at the time they create the MSS object.  Since each of these also has resources they will want to store in the object and reuse (shared memory buffers for XShmGetImage, offscreen Pixmaps and CUDA contexts for XComposite), that still seems reasonable.

When it comes time to add single-window capture, then that also will need to be chosen at the object's creation time: otherwise, the factory couldn't choose between XComposite and XShmGetImage.

Of course, a user could always keep multiple MSS objects around, and there's little downside to that.  But I'm not sure how you envision the API working, with regard to the scope of the MSS object.

BoboTiG:
For now, if "xlib" is kept as default, then users won't have anything to change. Later, if XCB becomes the default, we should use ShmGetImage since it seems the best for all cases (except onver SSH, in that case, we will update the documentation to tell users to use xlib).

Backends will be for users wanting more performances, and knowing what they want.

I do not see upcoming issues with this path, maybe am I missing something? We should not try to be too smart in the factory, but use the safest defaults "for the mass" I would say.

jholveck:
Yes, I do want the library to do the right thing without the user making choices.  MSS's ease of use is one of its greatest assets!

I think the factory can still make intelligent default choices without problems.  The decision between XShmGetImage and XComposite is pretty simple: capturing a single window to the GPU?  Use XComposite.  Otherwise?  Use XShmGetImage.  OBS Studio actually labels these two as "Single window" and "Whole desktop" (or something like that).

I've been planning on having the XShmGetImage backend automatically fall back to XGetImage if it can't use XShmGetImage.  It's difficult to detect in advance that you can't use XShmGetImage, but it's easy to switch if the call fails, and remember that decision.  My first prototypes testing XShmGetImage in MSS did just that.

So it should be easy to have the factory make intelligent default choices and "just work".  The user can specify a backend if they know of particular needs, but that'd be an advanced option, as you suggest.

I'm not worried about any of this being a problem.  I just wanted to make sure I understood your concept of the MSS object.  The user wouldn't be able to create one MSS object at the beginning and keep alternating between capturing to the CPU in one grab() call, the GPU in another, or do other things like that.  That's why I wanted to make sure the API would be okay with you.  It sounds like you're good with that concept, so I'll go ahead and finish up this code, and then start working on XShmGetImage.

jholveck:
Just to give you an update:

I'm working on a version that generates the XCB function calls from the XML definitions.  It's still pretty rough - I'm still simplifying it and cleaning things up - but you can see that on [jholveck:feat-xcb-codegen](https://github.com/jholveck/python-mss/tree/feat-xcb-codegen).  The code generator and (for the moment) the XML definitions are in [src/xcbproto](https://github.com/jholveck/python-mss/tree/feat-xcb-codegen/src/xcbproto), the generated code is in [src/mss/linux/xcbgen.py](https://github.com/jholveck/python-mss/blob/feat-xcb-codegen/src/mss/linux/xcbgen.py), and the MSS class that uses it is in [src/mss/linux/xgetimage.py](https://github.com/jholveck/python-mss/blob/feat-xcb-codegen/src/mss/linux/xgetimage.py).  As you can see, the "business logic" - the actual MSS implementation - is quite short and simple.

I'm trying to decide whether to go with this direction or not.  The code generator alone is longer than the generated code, or the hand-written version (I'm still simplifying the code generator, though).  But we can be confident that the generated code is correct, and not miscopied or something.  Also, as we add more X11 features (XShmGetImage, XComposite and Render, and later Present), we don't have to change the generator: we just say what additional functions we want to pull in.

We can check in the generated code, and even the protocol definitions, to ensure consistency.

Do you have a strong preference about whether we go with generated or hand-written code?  I think the generated version is probably going to be better long-term, but we can go either way.

BoboTiG:
The idea would be that we generare everything only once when we want to change anything in the C API, correct? That does not incur a performance penality when calling `python -m mss`, `mss`, or when using the module from another Python script?

The "generate" version is more code, but it will prevent any misuse on our side. It may improve future development speed also.
I see only good thing for now. Can you spot any potential downside?

jholveck:
> The idea would be that we generare everything only once when we want to change anything in the C API, correct? That does not incur a performance penality when calling `python -m mss`, `mss`, or when using the module from another Python script?

Exactly.

> The "generate" version is more code, but it will prevent any misuse on our side. It may improve future development speed also. I see only good thing for now. Can you spot any potential downside?

I agree.  It's just more code for now.

Some people just don't like generated code, or have particular ways they want it to work, so I was checking with you.  I'll proceed along that path!  Thanks!

BoboTiG:
Where do XML come from? Should we sync them from an upstream repository before calling the generator?

EDIT: `xproto.xml` seem to be from https://cgit.freedesktop.org/xcb/proto/tree/src/xproto.xml
We might want a small shell script to sync those file, but it's maybe too much since they won't change in years. Forget the sync part :)

jholveck:
That's the one, but its official location has moved (as noted in the [README in that repo](https://cgit.freedesktop.org/xcb/proto/tree/README.md)).  These days, the current location for that file is https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/blob/master/src/xproto.xml

But yes, that's the repo that all the XML files come from.  I intend to add relevant notes to the README about that.

It would be also possible to make the xorg/proto (where those XML files live) a git submodule instead of vendoring them in.  We really only need to update them if we're adding new features that get released in the future; the existing protocol specs are stable and forwards-compatible.  I thought it might be more trouble than it's worth to make it a submodule, but haven't really thought carefully about it yet.

BoboTiG:
Vendoring seems a good option, I guess.

jholveck:
One possibility, if you're uncomfortable with just copying the files in, would be to use [git subtree](https://www.atlassian.com/git/tutorials/git-subtree).  It can bring the xcbproto repo (which includes all the XML files we need) into the MSS repo, along with metadata to show the commit reference we've merged from, so we can pull in updates more easily.

Another alternative would be to use [git subrepo](https://github.com/ingydotnet/git-subrepo).  This is not yet an official part of git, but (from what I understand) was written by the author of git subtree to fix some of its shortcomings.  Notably, it includes metadata that shows the origin that the subtree came from.

Both of these give the tooling to manage the xcbproto repo as a subtree in the MSS repo, and without the complexity of git submodule.  The files will also be part of the MSS repo, so it'll be clear what versions we're using, without needing to configure git submodule.  But, these tools will bring in the entire xcbproto repo, when we only need a few files.

Since we'll rarely need to update these protocol definitions, and won't be making changes to them, I don't think it's necessary to use tooling like this.  But I can see the elegance of letting git know that these files come from somewhere different, if you'd prefer to.

BoboTiG:
Keep it simple: no need for more :)

jholveck:
I think this is good to go.  Take a look and see what you think.  There's little immediate user-facing improvement: just thread safety, and the new backend isn't the default in this patch.

Next, I'm working on adding the XShmGetImage backend, which will come with a performance improvement.  That will be much easier than the XCB change!

I'll be away next week, so you won't see that completed until after I get back.

src/mss/linux/xgetimage.py
```diff
@@ -0,0 +1,256 @@
...
+        self.drawable = self.root
+
+        # We don't probe the XFixes presence or version until we need it.
+        self._xfixes_ready: bool | None = None
+
+        # Probe the visuals (and related information), and make sure that our drawable is in an acceptable format.
+        # These iterations and tests don't involve any traffic with the server; it's all stuff that was included in the
+        # connection setup.  Effectively all modern setups will be acceptable, but we verify to be sure.
+
+        # Currently, we assume that the drawable we're capturing is the root; when we add single-window capture, we'll
```
BoboTiG, line 56: For now, let's keep it as-is. Maybe later could we add a toggle argument to run those tests or not (disabled by default for speed).

It could even be moved to a private method that one could call on purpose while debugging an issue, for instance.

How does that sound to you?

BoboTiG (approved):
Thank you very much, this is a huge step forward!

I love the inline documentation, very helpful.

🍾

src/mss/linux/xgetimage.py
jholveck: If you're talking about the comment on line 56, then I agree entirely that we don't need to change our assumption that we're only going to be capturing the root.  As far as that particular comment goes, I was mostly just duplicating the existing logic in the Xlib-based backend for this: it keeps self.root and self.drawable separate, which I guessed was done to support a future case when these might be different.

That assert is just so that we can assume that the drawable_depth and drawable_visual_id are the same as the root_depth and root_visual that we get with the initial setup.  These are defined to be the same on line 46.  If the root and drawable could be different, then we'd have to issue a new GetProperties call on the drawable.  If, at some point, we decide to make these different, then that assert will fire and help us remember to update lines 59-60.

In other words, that comment isn't meant to apply to all of the following lines.  We can add a blank line before line 61 to make that more clear.

But if you're talking about the remaining tests - the bulk of `__init__` - then I'll talk about that.

These tests are ensuring that the format is exactly what we expect: that there are 32 bits per pixel, they correspond to RGB values (instead of being indexes into a colormap), in BGRX order (bytewise), with no additional padding on each line.  This is the normal format for modern systems, including anything I've found with 24/32-bit visuals.  It's what MSS had assumed already; we're just testing these assumptions to ensure that we will alert the user if there's an extremely oddball setup thrown into the mix.

These tests are very fast.  They don't require any additional round-trips with the server, since they're just examining information that was automatically exchanged when the connection was established (as part of the X11 setup protocol).  They also only happen when the MSS instance is created, rather than at each call to grab().  If somebody is concerned about speed, then they won't be creating a new MSS instance for each grab, as the docs point out.

We can skip these tests, but that might come with problems: the user doesn't have any information about how to interpret the data they're getting.  They don't know which channel is which, and don't even know if these are RGB values or if they're indexes into an inaccessible colormap.  You've seen related issues (such as #207) about providing the consumer with information about how to interpret the pixel data.

I would be okay with adding attributes to the ScreenShot object that indicate how the raw data should be interpreted.  As one example, I'd actually like to add a new attribute that indicates whether there's a genuine alpha channel: in the existing Linux implementations, the "alpha" channel, if you're capturing an opaque window, is often all zeros - which means the entire area is transparent.  I've also thought that it would be good to tag it with the colorspace (again, see #207), but haven't really worked out the best API for that.

But, for simplicity for the end users, I felt that it would be best to verify that the data being supplied matches the general expectation: a BGRX 32-bpp layout.

If the user has prior knowledge about the pixel layout - such as knowing that there's some padding on each row, or that the machine is big-endian - and have been accommodating that, then these extra checks may mean that it will throw an exception where there wasn't one before.  But I don't see this as a realistic regression.

That's my reasoning for doing more careful testing about the format we'll be providing.  It's fast, and it means that the user can know for sure what the data means.  There is an argument to be made that this prevents users from getting screenshots in unusual formats, if they know to expect it.  But, I think that we should give the user information about the pixel layout, convert it automatically, or raise an exception.

That's my reasoning, but if you prefer, we can certainly take out these tests.  They aren't required for the work I'm doing with MSS; I just thought that it would be the best way to proceed for the wide variety of systems that other MSS users might use.

jholveck:
Terrific; I'm glad you like it!

If you have additional changes you'd like, then of course you can add them on main.  I'm just glad we could get this in place, so we can move forward with some more user-visible changes!

src/mss/linux/xgetimage.py
BoboTiG: You made a point. Lets keep it as-is then.

PR #431, opened 2025-11-22
Implement an XShmGetImage-based backend

### LINKED ISSUE #96: Improve speed on GNU/Linux with XShm
XShm seems a good use for improving speed on GNU/Linux.

See https://stackoverflow.com/a/43500537/1117028 for a idea of implementation.

We could also find valuable information here: http://www.rahul.net/kenton/perf.html

The goal is to let MSS find the best implementation to use at initialization

jholveck:

Following on to #426's introduction of XCB backends, this change introduces a backend based on XShmGetImage (technically, xcb_shm_get_image).  This avoids the need to send the pixel data across the socket.  Instead, the X11 server saves the pixel data into a memory buffer that's shared between the server and MSS, and MSS can use that buffer directly.

This is quite performant: on my computer, the new backend can take 4k screenshots at 30-34 fps, while the XGetImage backend could only run at 11-14 fps.

This backend requires the client and server to be on the same machine (so they can share a memory buffer).  It often can't be used across separate containers on the same machine (depending on the container configuration).  Because of these limitations, the existing XGetImage implementation is still available.  It can be explicitly requested by the user, or the new implementation will use it automatically if XShmGetImage isn't possible.

- Fixes #96 

- [x] Tests added/updated
- [x] Documentation updated

Is your code right?

- [x] `./check.sh` passed

### DIFF
--- a/src/tests/test_implementation.py
+++ b/src/tests/test_implementation.py
@@ -95,72 +95,104 @@ def test_factory_unknown_system(backend: str, monkeypatch: pytest.MonkeyPatch) -
     assert error == "System 'chuck norris' not (yet?) implemented."
 
 
-@patch.object(sys, "argv", new=[])  # Prevent side effects while testing
+@pytest.fixture
+def reset_sys_argv(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(sys, "argv", [])
+
+
+@pytest.mark.usefixtures("reset_sys_argv")
 @pytest.mark.parametrize("with_cursor", [False, True])
-def test_entry_point(with_cursor: bool, capsys: pytest.CaptureFixture) -> None:
-    def main(*args: str, ret: int = 0) -> None:
+class TestEntryPoint:
+    """CLI entry-point scenarios split into focused tests."""
+
+    @staticmethod
+    def _run_main(with_cursor: bool, *args: str, ret: int = 0) -> None:
         if with_cursor:
             args = (*args, "--with-cursor")
         assert entry_point(*args) == ret
 
-    # No arguments
-    main()
-    captured = capsys.readouterr()
-    for mon, line in enumerate(captured.out.splitlines(), 1):
-        filename = Path(f"monitor-{mon}.png")
-        assert line.endswith(filename.name)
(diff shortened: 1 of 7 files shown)


BoboTiG:
Tested: it works fine, and is indeed faster than the Xlib backend :muscle:

jholveck:
I'm working on the user docs, so we should talk about how we want to present this.

Earlier, I think we considered making a release with the new XCB-based stuff experimental and non-default, to give advanced users a chance to try it out and report issues.  Do you think that's what you want to do?  The alternative I see is to make XShmGetImage the default backend.  I think it's safe to make the new XCB-based XShmGetImage the default, but it's up to you.

Once I write the docs, I don't think there's anything else we need to do before committing it, and then you can make a release.  What do you think?

BoboTiG:
> (...) The alternative I see is to make XShmGetImage the default backend. I think it's safe to make the new XCB-based XShmGetImage the default, but it's up to you.

I think too, lets use XShmGetImage as the default backend right now.
 
> Once I write the docs, I don't think there's anything else we need to do before committing it, and then you can make a release. What do you think?

Maybe adding a CLI argument to be able to choose the backend would be interesting?

jholveck:
I've changed the default to use XShmGetImage on Linux, and added a `--backend` flag.  Let me know if there's anything else you come across during your review!

src/tests/test_implementation.py
```diff
@@ -95,72 +95,104 @@ def test_factory_unknown_system(backend: str, monkeypatch: pytest.MonkeyPatch) -
...
+        backend = "default"
+        for opt in ("-b", "--backend"):
+            self._run_main(with_cursor, opt, backend, "-m1")
+            captured = capsys.readouterr()
+            filename = Path(captured.out.rstrip())
+            assert filename.is_file()
+            filename.unlink()
+
+    def test_invalid_backend_option(self, with_cursor: bool, capsys: pytest.CaptureFixture) -> None:
+        backend = "chuck_norris"
```
BoboTiG, line 191: 😄

BoboTiG:
That's terrific, thank you @jholveck 🍾

Issue #442, opened 2025-12-21
mss.grab() is blocking indefinitely after a certain amount of usage


cjshrader:
Hey there,

I maintain an open source program that helps with Diablo 4 loot filtering. As part of this, we read the screen using mss very often. We are Windows only.

After 30 minutes to an hour, the program will freeze up entirely. I've been able to trace where it's getting stuck to this code in mss's base.py:

```python
        with lock:
            screenshot = self._grab_impl(monitor)
            if self.with_cursor and (cursor := self._cursor_impl()):
                return self._merge(screenshot, cursor)
            return screenshot
```

Specifically on the `with lock:` step. 

I will go through all of our code in case it is beneficial, but at a high level the user presses a hotkey which kicks off a thread. That thread will read the screen to see if a certain menu is open. It's at that point that everything gets blocked, with no errors or any interaction. A hotkey to kill that thread works fine, but running it again and every time after that will always be blocked.

I believe we are calling mss correctly:

```python
        with mss.mss() as sct:
            img = np.array(sct.grab(self.window_roi))
```

We then perform handling on the image that we receive.

Here is the full flow with lines from our code in case any of that is beneficial:

First, the user presses a hotkey. We'll use for example the hotkey to move items from inventory to stash. It kicks off a thread in main, seen here:

https://github.com/d4lfteam/d4lf/blob/27630ecbf3d121cc4253dea815f10251f5cc71b5/src/scripts/handler.py#L72

```python
                    self.loot_interaction_thread = threading.Thread(
                        target=self._wrapper_run_loot_interaction_method,
                        args=(loot_interaction_method, method_args),
                        daemon=True,
                    )
                    self.loot_interaction_thread.start()
```

Which I will shortcut some of this, but it calls loot_mover.py (https://github.com/d4lfteam/d4lf/blob/27630ecbf3d121cc4253dea815f10251f5cc71b5/src/loot_mover.py#L23) which calls stash.is_open() (https://github.com/d4lfteam/d4lf/blob/27630ecbf3d121cc4253dea815f10251f5cc71b5/src/ui/menu.py#L55) which calls detect (https://github.com/d4lfteam/d4lf/blob/27630ecbf3d121cc4253dea815f10251f5cc71b5/src/template_finder.py#L72) which calls Cam().grab() which is where we call mss:

https://github.com/d4lfteam/d4lf/blob/27630ecbf3d121cc4253dea815f10251f5cc71b5/src/cam.py#L79

```python
[...]

PR #443, opened 2025-12-21
Use Sphinx autodoc for documentation.

jholveck:
This changes the API documentation generation to use Sphinx autodoc.  The existing API docs were moved into docstrings.

There's still cleanups to be done on the documentation, but this seemed like a place to start.

My change does delete some imported classes and constants that were in the API docs for reasons I'm not really clear on, such as mss.darwin.CGPoint and mss.windows.MONITORNUMPROC.  Should these still be in the docs?

### DIFF
--- a/src/mss/models.py
+++ b/src/mss/models.py
@@ -20,7 +20,7 @@ class Pos(NamedTuple):
 
 
 class Size(NamedTuple):
-    #: The hoirizontal X width.
+    #: The horizontal X width.
     width: int
     #: The vertical Y height.
     height: int


jholveck:
I suspect that the errors are about the changes to how close() is handled, but I'll investigate.  I can reproduce them here with xvfb-run.

BoboTiG:
Nice!

> My change does delete some imported classes and constants that were in the API docs for reasons I'm not really clear on, such as mss.darwin.CGPoint and mss.windows.MONITORNUMPROC. Should these still be in the docs?

No, that's OK to document only objects that users will encounter.

src/mss/base.py
```diff
@@ -91,17 +112,43 @@ def _monitors_impl(self) -> None:
...
+
+        This will be called at most once.
+        """
+        # It's not necessary for subclasses to implement this if they have nothing to clean up.
+
+    def close(self) -> None:
+        """Clean up.
+
+        This releases resources that MSS may be using.  Once the MSS
+        object is closed, it may not be use used again.
```
BoboTiG:
```suggestion
        object is closed, it may not be used again.
```

src/mss/base.py
```diff
@@ -91,17 +112,43 @@ def _monitors_impl(self) -> None:
...
+        It is safe to call this multiple times; multiple calls have no
+        effect.
+
+        Rather than use :py:meth:`close` explicitly, we recommend you
+        use the MSS object as a context manager::
+
+            with mss.mss() as sct:
+                ...
+        """
+        with lock:
```
BoboTiG, line 136: This flow seems incorrect, right?

I would suggest something like that to not using the lock for nothing:

```python
if self._closed:
    return

with lock:
    self._close_impl()
    self._closed = True
```

jholveck: With the suggested flow, you can have two different threads both try to run the close at the same time.

Example:
1. Thread A runs `close()`.  It seems that `_closed` is false, and starts running `_close_impl()`.
2. While A is inside `_close_impl()`, thread B runs `close()`.  It checks `_closed` **before acquiring the lock**, and also sees False.
3. A finishes, sets `_closed = True`, and releases the lock.
4. B now acquires the lock, and runs `_close_impl()` a second time.

To prevent this, the "have we already closed?" check has to be inside the lock. Even though the true path just returns, we still need mutual exclusion so that checking `_closed` and committing to performing the close are atomic with respect to other callers.

I also realize that I'm making a new guarantee in this documentation: that it's safe to call `close()` more than once.  I thought that adding this check would be useful for new users, who might be unaccustomed to `with:` semantics, and might try to close the object as well.

That said, as I think about it, I think that this test would probably be better in a per-object lock, than using the global lock.  The same might apply to `grab()`, which I note also acquires the global lock.  But I figured that's a change that's more involved than I wanted to fold into a documentation change PR.

BoboTiG: About the per-object lock, maybe this would be helpfull in this scenario: #442. But this is just a guess, it would be cleaner anyway.

jholveck: Yeah, that's a reasonable guess.  We can deal with moving to a per-object lock separately; I'll open an issue about that.

At a first glance at the issue, I don't see anything blatant that would cause a deadlock.  However, I do see that their example of calling into MSS creates a new MSS object each time they want to take a screenshot.  Moving to per-object locks, where they're possible, would help with those cases.

This may require an audit of the existing `_grab_impl` code, to make sure the implementations don't have any implicit assumptions that the global lock is held.  If they do require that no other MSS objects are doing a screen grab at the same time, then there would be a need for them to use a global lock.

BoboTiG:
Should we merge this, and iterate if needed?

src/mss/darwin.py
BoboTiG: It would be interesting to have this one documented.

BoboTiG, line 108: Do you know a way to have their values and types printed in the docs (this apply to windows and linux class attrs as well)?

src/mss/models.py
BoboTiG, line 17: Note for later: document those fields to not have useless details in the docs like "Alias for field number 0".

src/mss/screenshot.py
```diff
@@ -63,33 +73,46 @@ def from_size(cls: type[ScreenShot], data: bytearray, width: int, height: int, /
...
-        """BGRA values from the BGRA raw pixels."""
-        return bytes(self.raw)
+        """BGRx values from the BGRx raw pixels.
 
-    @property
-    def height(self) -> int:
-        """Convenient accessor to the height size."""
-        return self.size.height
+        The format is a bytes object with BGRxBGRx... sequence.  A specific
+        pixel can be accessed as bgra[(y * width + x) * 4:(y * width + x) * 4 + 4].
```
BoboTiG:
```suggestion
        pixel can be accessed as ``bgra[(y * width + x) * 4:(y * width + x) * 4 + 4]``.
```

src/mss/screenshot.py
```diff
@@ -63,33 +73,46 @@ def from_size(cls: type[ScreenShot], data: bytearray, width: int, height: int, /
...
+            msg = f"Pixel location ({coord_x}, {coord_y}) is out of range."
+            raise ScreenShotError(msg) from exc
+
     @property
     def rgb(self) -> bytes:
         """Compute RGB values from the BGRA raw pixels.
 
-        :return bytes: RGB pixels.
+        The format is a bytes object with BGRBGR... sequence.  A specific
+        pixel can be accessed as rgb[(y * width + x) * 3:(y * width + x) * 3 + 3].
```
BoboTiG:
```suggestion
        pixel can be accessed as ``rgb[(y * width + x) * 3:(y * width + x) * 3 + 3]``.
```

src/mss/darwin.py
jholveck: You can put comments starting with `#:` just before where they're set in `__init__` to describe them.  Similarly, you can put a type declaration where they're set.

jholveck:
> Should we merge this, and iterate if needed?

Yeah, that's what I planned last night... then came across that last batch of doc fixes I added to the PR!

Later today, I can incorporate your suggestions, check the doc build, and mark the PR as ready.

src/mss/models.py
```diff
@@ -14,10 +13,14 @@
...
 
 class Pos(NamedTuple):
+    #: The horizontal X coordinate of the position.
     left: int
+    #: The vertical Y coordinate of the position.
     top: int
 
 
 class Size(NamedTuple):
+    #: The hoirizontal X width.
```
BoboTiG:
```suggestion
    #: The horizontal X width.
```

BoboTiG:
Thank you very much @jholveck, that's cleaner now, and properly documented from the code. I love it!

PR #444, opened 2025-12-21
Add demos

jholveck:

Adds demos, which are longer examples than belong in the regular documentation.  The first of these is one that streams to a toy TV.  Before merging this, I'd like to get a sense of how much of the code would be shared with the other demo I've got in progress, one that saves to an AVI file.

- [ ] Tests added/updated
- [ ] Documentation updated

Is your code right?

- [ ] `./check.sh` passed

### DIFF
--- a/demos/tinytv-stream.py
+++ b/demos/tinytv-stream.py
@@ -736,13 +736,15 @@ def show_stats(byte_counts: Iterable[int]) -> None:
 
     :param byte_counts: Iterable of byte counts per frame.
     """
-    start_time = time.clock_gettime(time.CLOCK_MONOTONIC)
+    # If we needed high-precision, such as for benchmarking very short times, we might want to use time.perf_counter().
+    # However, time.monotonic() is sufficient for simple stats reporting.
+    start_time = time.monotonic()
     time_deque: deque[float] = deque(maxlen=100)
     byte_count_deque: deque[int] = deque(maxlen=100)
     next_display_update = 0.0
     last_status_len = 0
     for frame_count, byte_count in enumerate(byte_counts):
-        now = time.clock_gettime(time.CLOCK_MONOTONIC)
+        now = time.monotonic()
         time_deque.append(now)
         byte_count_deque.append(byte_count)
         if now >= next_display_update and len(time_deque) > 1:


demos/tinytv-stream.py
```diff
@@ -0,0 +1,1005 @@
...
+    per-frame decoding in the TinyTV, rather than the raw transfer
+    speed, is the limiting factor.
+
+    This is run on the main thread.  This is partly a matter of
+    convenience, and partly because it simplifies waiting for the
+    pipeline to complete.
+
+    :param byte_counts: Iterable of byte counts per frame.
+    """
+    start_time = time.clock_gettime(time.CLOCK_MONOTONIC)
```
BoboTiG: Would it be a good idea to go with `time.monotonic()`?

This would make it work on Windows as well.

jholveck:
Thank you!  I'm glad you like it!  MSS is popular enough that it's gone beyond the simple "take a screenshot, save a PNG" use cases.  I think that having clear demos showing how to use a library in practical situations, beyond simple examples, is a big help.

As for the code reuse question, this is something I'm of two minds about.  On one hand, yes, it's great for demos to be self-contained.  On the other hand, I've also got the video encoding demo (you can see [an early test](https://github.com/jholveck/python-mss/blob/feat-demo-video/demos/video-capture.py), although that's using some features that I was playing with in a separate branch).  That uses the same basic pipeline concept.  It's still "input -> processing -> output" in separate threads.

The pipeline code in the tinytv-stream demo is nearly 300 lines long.  I was thinking it might be worth separating that into a separate file, rather than copying it into a separate 

That would let users review both demos without having to check to see if all that code is the same or if they need to re-read it all.  It would also make sure that we don't accidentally change the pipeline code in one and not the other.

So as I said, I'm conflicted.

That said, I think we're fine committing this as-is, and possibly factoring things out later if we want to.

BoboTiG:
Ah yes, the pipeline code could be in its own file: it will reduce demos code, and ease understanding MSS-specific usage.

jholveck:
Ok!  Do you want me to do that before we commit these, or as part of the video streaming demo commit?

jholveck:
By the way, while writing this I was wondering if it was a particularly useful demo, since it's so focused on pipelines and such, and not much about MSS.  Then I realized that there's a reason for that: MSS is so easy to use that it doesn't require a lot of support code!

demos/tinytv-stream.py
jholveck: Absolutely; thanks for catching that!

BoboTiG:
That's insanely usefull, we keep it! :)

BoboTiG:
A link in the documentation can be added later, I'm now hungry to see what other cool stuff your wrote with MSS :D

PR #445, opened 2025-12-21
Add a test for tuples that have negative coordinates.

jholveck:
Specifically, a negative width or height.

Particularly, this seems like an easy case for a user to hit if somebody forgets that the PIL tuple style uses top, left, RIGHT, BOTTOM and not top, left, WIDTH, HEIGHT.

Ask how I know.

Test that we do handle that gracefully.

- Quick test for a case I hit today.

- [X] Tests added/updated

Is your code right?

- [X] `./check.sh` passed

### DIFF
--- a/src/tests/test_gnu_linux.py
+++ b/src/tests/test_gnu_linux.py
@@ -170,39 +170,6 @@ def test_unsupported_depth(backend: str) -> None:
         sct.grab(sct.monitors[1])
 
 
-def test_region_out_of_monitor_bounds(display: str, backend: str) -> None:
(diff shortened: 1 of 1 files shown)


BoboTiG:
I see macOS and Windows are less strict.

Should we add checks in our code instead of relying on the backend to fail?
It would also properly fix issues like in #270: invalid region or region with size < 1 should be reported early.

jholveck:
Yeah, I was surprised about that here too!

I'll add a test in grab to check the bounds, and make sure it gets invoked in the other paths too.

PR #448, opened 2026-01-03
Add error-checking to Windows implementation

jholveck:

Added error checking to the Windows system calls.  Also cleaned up a few of the types we have in the signatures, but nothing that would have had an adverse effect.

I can add some testing, such as by adding 1 to `sct._handles.memdc` between when we create the object and when we grab a screenshot.  Really, there aren't many black-box ways to provoke a Windows error in automated testing that I can think of.

- [ ] Tests added/updated
- [x] Documentation updated  (N/A)

Is your code right?

- [x] `./check.sh` passed

### DIFF
--- a/src/mss/models.py
+++ b/src/mss/models.py
@@ -1,7 +1,7 @@
 # This is part of the MSS Python's module.
 # Source: https://github.com/BoboTiG/python-mss.
 
-from typing import Any, Callable, NamedTuple, Optional
+from typing import TYPE_CHECKING, Any, Callable, NamedTuple
 
 Monitor = dict[str, int]
 Monitors = list[Monitor]
@@ -9,8 +9,9 @@ Monitors = list[Monitor]
 Pixel = tuple[int, int, int]
 Pixels = list[tuple[Pixel, ...]]
 
-CFunctions = dict[str, tuple[str, list[Any], Any]]
-CFunctionsErrChecked = dict[str, tuple[str, list[Any], Any, Optional[Callable]]]
+if TYPE_CHECKING:
+    CFunctions = dict[str, tuple[str, list[Any], Any]]
+    CFunctionsErrChecked = dict[str, tuple[str, list[Any], Any, Callable | None]]
 
 
 class Pos(NamedTuple):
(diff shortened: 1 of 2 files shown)


BoboTiG:
Nice! A note in the changelog and we will be good to merge.

No need to bother with a test.

It will help looking for the real root cause sometimes when a generic error message would have been raised 👍

src/mss/models.py
```diff
@@ -10,7 +10,7 @@
 Pixels = list[tuple[Pixel, ...]]
 
 CFunctions = dict[str, tuple[str, list[Any], Any]]
-CFunctionsErrChecked = dict[str, tuple[str, list[Any], Any, Callable | None]]
+CFunctionsErrChecked = dict[str, tuple[str, list[Any], Any, Optional[Callable]]]
```
BoboTiG: Why is Optional needed here? The union character is better IMO.

jholveck: I agree, the union is better style.  But Python 3.9 apparently doesn't like it; see the first check-in test.

I guess elsewhere, usually we're normally only using unions in annotations, where it's only actually evaluated when we run MyPy, which we only do with 3.12.  But here, we're defining a new type, which is outside a type annotation.  In that case, Python tries to evaluate the code, not knowing for sure yet if it's a regular variable or what.

Or something like that.

We could instead put that (and the corresponding import) in a TYPE_CHECKING guard.  That might be cleaner stylistically.

To be honest, I didn't dig into that error (the one from the check-in test) as much as I normally would.

I wanted to get the error-checking in quickly so that we could give the user in #442 a test to rule out problems with how we call Windows functions.  Previously, we only had one line where we checked for an error; if any of the other Win32 API calls had failed, who knows how things would play out.

I couldn't think of any conditions where that seemed likely, but it seemed worth ruling out.

BoboTiG: Talking about such issue, I guess this is fine dropping Python 3.9 support:
- it's EOL since the last year (https://devguide.python.org/versions/)
- it represents ~3% of total downloads (https://pypistats.org/packages/mss)

jholveck: I went ahead and put the CFunctions and CFunctionsErrChecked in a TYPE_CHECKING guard, and switched back to using `| None` instead, to keep things going until you're ready to officially drop 3.9 support in a separate commit.  I also added a changelog entry,

PR #452, opened 2026-01-04
Use per-object locking instead of global locking

### LINKED ISSUE #446: Move to per-object locks
Presently, code that requires locks uses the course-grained, global mss.lock.  Many of these cases could be moved to per-object locks, if they only require locking over one MSS object's structure, instead of the underlying OS or other global resources.  This is suspected to help with #442.  It was also mentioned in #443, which started using the global lock to protect against closing an object multiple times.

(Feel free to assign this to me.  I just didn't want to forget about it over the holidays!)

jholveck:
The global lock just provides too much of a surface for contention and deadlocks.  Global locking is still needed by the Xlib backend, since Xlib is not thread-safe and we're not enabling the parts that make it slightly more thread-safe, so it now has its own lock object.

The docs were also updated to spell out our multithreading guarantees.

I'm marking this PR as a draft because I want to review it more closely, rather than submitting a quick bit of weekend code.

- Fixes #446

- [ ] Tests added/updated
- [x] Documentation updated

Is your code right?

- [x] `./check.sh` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ See Git checking messages for full history.
 ## 10.0.0 (2024-11-14)
 - removed support for Python 3.8
 - added support for Python 3.14
-- Linux: fixed a threadding issue in `.close()` when calling `XCloseDisplay()` (#251)
+- Linux: fixed a threading issue in `.close()` when calling `XCloseDisplay()` (#251)
 - Linux: minor optimization when checking for a X extension status (#251)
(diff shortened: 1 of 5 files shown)


BoboTiG:
Thank you @jholveck 🥂 

That looks good for a first check on my end 👍 

(Think about the changelog entry ;))

jholveck:
Sorry I keep forgetting to include the CHANGELOG entry.  Thanks for keeping everything clean and documented!

PR #453, opened 2026-01-04
Fix object release in macOS

### LINKED ISSUE #450: Investigate object ownership in macOS
Currently, in darwin.py's `_grab_impl`, we release `prov`, and not `image_ref`.  I think those are mistaken.  (Our use of `copy_data` seems correct.)

CoreFoundation (including CoreGraphics) has [specific rules about object ownership](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-CJBEJBHH).  Under those rules, if you get an object through a function with `Create` or `Copy` in the name, you own it, and are responsible for releasing it with `CFRelease` or similar.  If you get an object through a function with `Get` in the name, you do _not_ own it: you can retain/release it if you need to hold it long-term.

* We got `image_ref` from a Create function, so we do own it.  That means we should call `CFRelease` (or `CGImageRelease`, a tiny wrapper) on it.
* We got `prov` from a Get function, so don't own it.  That means that we shouldn't call `CGDataProviderRelease` (a tiny wrapper around `CFRelease`) on it, since we're only using a borrowed reference.

This should be investigated and fixed if necessary.

jholveck:
Previously, we would release the image data provider, but not the image.  We actually need to release the image, and not the data provider.  This is based on the CoreFoundation "Create/Copy/Get" ownership rules: we own things that we allocate with functions with Create or Copy in their name, but not things that have Get in their name.

I suspect that the CGImage held a reference to its data provider, which stored the pixel data.  When we freed the data provider, we released the pixel data, but not the CGImage structure.

Now, we're releasing the CGImage, and it releases the data provider.

Empirical testing while taking 10000 successive screenshots showed that, once steady state was reached, the previous version was leaking an average of about 467 bytes per screenshot on average.  The new version still leaks about 148 bytes per screenshot, although that may not be related to macOS.

Also fixed a few function types, but those were harmless.

- Fixes #450

- [ ] Tests added/updated  (N/A)
- [ ] Documentation updated  (N/A)

Is your code right?

- [x] `./check.sh` passed

### DIFF
--- a/src/mss/darwin.py
+++ b/src/mss/darwin.py
@@ -90,7 +90,6 @@ CFUNCTIONS: CFunctions = {
     "CFDataGetBytePtr": ("core", [c_void_p], POINTER(c_ubyte)),
     "CFDataGetLength": ("core", [c_void_p], c_long),
     "CFRelease": ("core", [c_void_p], None),
-    "CGDataProviderRelease": ("core", [c_void_p], None),
     "CGGetActiveDisplayList": ("core", [c_uint32, POINTER(c_uint32), POINTER(c_uint32)], c_int32),
     "CGImageGetBitsPerPixel": ("core", [c_void_p], c_size_t),
     "CGImageGetBytesPerRow": ("core", [c_void_p], c_size_t),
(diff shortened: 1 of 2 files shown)


src/mss/darwin.py
```diff
@@ -219,10 +230,9 @@ def _grab_impl(self, monitor: Monitor, /) -> ScreenShot:
                     cropped.extend(data[start:end])
                 data = cropped
         finally:
-            if prov:
-                core.CGDataProviderRelease(prov)
```
BoboTiG, line 223: Then the line 93 can be removed since the function is no more used.

BoboTiG:
Thanks @jholveck!


A line in the changelog might be interesting?

PR #456, opened 2026-01-17
New demo: video encoding

jholveck:

It seems that there are a number of users who use MSS for video encoding.  Add a demo showing how to do this.

This again highlights multithreaded pipelining, like the TinyTV demo, but is more accessible (as few users have a TinyTV).  It also shows a number of pitfalls that are common when encoding video, such as failing to correctly record timestamps.

- [ ] Tests added/updated  - N/A
- [x] Documentation updated
- [x] Changelog entry added
- [ ] `./check.sh` passed

### DIFF
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -49,7 +49,8 @@
 # Timestamps (PTS/DTS)
 # --------------------
 #
-# Every frame has a *presentation timestamp* (PTS): when the viewer should see it.
+# Every frame has a *presentation timestamp* (PTS): when the viewer should see it.  (See the next section for how
+# these are represented.)
 #
 # Encoders may output packets in a different order due to B-frames.  Those packets also have a *decode timestamp*
 # (DTS): when the decoder must decode them so the PTS schedule can be met.
@@ -73,8 +74,12 @@
 # Many video files run at a fixed frame rate, like 30 fps.  Each frame is shown at 1/30 sec intervals.  This is called
 # *constant frame rate*, or *CFR*, and that's what we do in the simple version of this demo.
 #
-# One problem with this is that, if the encoder can't keep up, the video will appear sped-up when played back.  The
-# comments at the beginning of the simple version of this demo go into more detail about that problem.
+# Applications using CFR usually set the time base to the frame rate, such as 1/30 sec.  This lets them just use the
+# frame number for the PTS.
+#
+# One problem with real-time recording to CFR is that, if the encoder can't keep up, the video will appear sped-up
+# when played back.  The comments at the beginning of the simple version of this demo go into more detail about that
+# problem.
 #
 # In this advanced version, we use *variable frame rate*, or *VFR*.  That's because we can't be sure that the encoder
 # will be able to work fast enough: we haven't tuned its settings for your screen resolution and hardware.  While the


BoboTiG:
Nice!!

BoboTiG:
This is really good information here, priceless!

If I wanted to use h265, is it as easy as serting "h265", or does it need special handling?

jholveck:
You could pass `--codec libx265` on the command line.  The thing you use there is the same as ffmpeg's `-c:v` flag.

You can get a list of available codecs in the PyAV build using `python3 -m av --codecs`; libx265 is among those.

You'd need to comment out the `"profile":"high"` in `CODEC_OPTIONS`: libx265 doesn't recognize the high profile.  Most, if not all, of the features in the H.264 "high" profile are already part of the H.265 "main" (default) profile.

You can look at other flags for libx265 using `ffmpeg --help encoder=libx265`, if your ffmpeg build has libx265 compiled in.  The libx265 library is GPL-only, so some builds might not include it, but the one included in PyAV does.

I'll add some comments to this effect.

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+        monitor = sct.monitors[1]
+
+        with av.open(FILENAME, "w") as avmux:
+            # The "avmux" object we get back from "av.open" represents
+            # the MP4 file.  That's a container that holds the video,
+            # as well as possibly audio and more.  These are each
+            # called "streams".  We only create one stream here, since
+            # we're just recording video.
+            video_stream = avmux.add_stream(
+                CODEC, rate=FPS, options=CODEC_OPTIONS
```
bboudaoud-nv: This will encode video as pure CFR (with the time base equal to 1/rate for some encoders) if you'd like you could demonstrate setting the time base here instead then encode the frame time of each frame with the capture below to account for variable timing (VFR).

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+            # time that is spent doing the capture and encode (which
+            # can be substantial) is counted as part of the total time
+            # we need to delay.
+            next_frame_time = capture_start_time
+
+            print("Capturing to", FILENAME, "for", CAPTURE_SECONDS, "seconds")
+            while True:
+                # Wait until we reach the time for the next frame.
+                while (now := time.monotonic()) < next_frame_time:
+                    time.sleep(next_frame_time - now)
```
bboudaoud-nv, line 103: Sleeping this full time first means that the capture will always be at least a bit slower than the target rate. What you could do is capture a loop start time here, perform your capture, then sleep until the next frame time so that you account for any variability in the `sct.grab()` call below.

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+                # from_numpy_buffer.  But the most obvious way is to
+                # use PIL: you can create an Image from the
+                # screenshot, and create a VideoFrame from that.  That
+                # said, if you want to boost the fps rate by about
+                # 50%, check out the advanced demo, and search for
+                # from_numpy_buffer.
+                img = Image.frombytes(
+                    "RGB", screenshot.size, screenshot.bgra, "raw", "BGRX"
+                )
+                frame = av.VideoFrame.from_image(img)
```
bboudaoud-nv, line 128: An alternative, faster route, especially if MSS can capture in BGR is to assign video planes directly here. You can avoid the Pillow Image creation time and instead use that to directly assign a `frame.planes[0:2]` to your BGR data here (as encoded bytes from the arrays)

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+                # this call are whatever the encoder is ready to give
+                # you, not necessarily the packets related to the
+                # frame you're handing it right now.
+                packets = video_stream.encode(frame)
+
+                # As we said, the MP4 file is a bunch of packets from
+                # possibly many streams, all woven (or "muxed")
+                # together.  So the ultimate destination of the data
+                # is to send it to the MP4 file, avmux.
+                avmux.mux(packets)
```
bboudaoud-nv, line 154: This may be fine, but I've always used:

```python
for packet in stream.encode(frame):
  avmux.mux(packet)
```

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+            print()
+
+            # Earlier, we mentioned that the encoder might hold onto
+            # some frames, while it decides how to encode them based
+            # on future frames.  Now that we're done sending it
+            # frames, we need to get the packets for any frames it's
+            # still holding onto.  We do this by sending None instead
+            # of a frame object.
+            packets = video_stream.encode(None)
+            avmux.mux(packets)
```
bboudaoud-nv, line 164: Typically referred to as "flushing" the stream. Again in their docs av does:
```python
for packet in stream.encode():
  avmux.mux(packet)
```

demos/video-capture-simple.py
```diff
@@ -0,0 +1,225 @@
...
+
+# Within an MP4 file, the video can be stored in a lot of different
+# formats.  In this demo, we use H.264, since it's the most widely
+# supported.
+#
+# In ffmpeg, and the av libraries that we use here, the best codec for
+# H.264 that doesn't require any specific hardware is libx264.  There
+# are faster ones that are hardware-accelerated, such as h264_nvenc
+# which uses specialized chips on Nvidia video cards.
+CODEC = "libx264"
```
bboudaoud-nv, line 61: I believe this (and affiliated codec options) are the only reason this script wouldn't also run on Windows. You could make these platform dependent, or support them and the filename/capture time/frame rate as arguments?

bboudaoud-nv:
> This is really good information here, priceless!
> 
> If I wanted to use h265, is it as easy as serting "h265", or does it need special handling?

@BoboTiG codec names can be a bit finicky here. IIRC on Linux `x265` should work for encode. On Windows I believe its `hevc`. The optional parameters passed in, as well as the supported encoding formats for frames may change with the codec depending on the system though. Currently h264 encoding is definitely the safe choice for multi-platform.

You could also experiment with codec names like `mpeg` or `h264` to see if you get better multiplatform support here.

BoboTiG:
> Currently h264 encoding is definitely the safe choice for multi-platform.

Fully agree on that.

Thank you for the useful review :)

BoboTiG:
@jholveck should we do checks for third-party modules?

I was testing, and I was missing multiple modules, I needed to look on PyPI, a process which could be improved.

Should we check at import time, something like this?

```
try:
    import av
except ImportError:
    print("The PyAv module is missing, run: `python -m pip install av`")
    sys.exit(1)

try:
    import si_prefix
except ImportError:
    print("The si-prefix module is missing, run: `python -m pip install si-prefix`")
    sys.exit(1)
```

This feels like a horrible solution haha, I'm simply putting words on a idea.

BoboTiG:
Overall, it works great! I see a small difference in colors, on Linux, but I guess this is due to the JPEG compression.

jholveck:
> Overall, it works great! I see a small difference in colors, on Linux, but I guess this is due to the JPEG compression.

You might try turning on DISPLAY_IS_SRGB.

demos/video-capture-simple.py
jholveck: Isn't that exactly what we do?  next_frame_time starts at the current time (so no sleep), and then is incremented by 1/30 sec each iteration.

demos/video-capture-simple.py
jholveck: Yes, see the comment above this bit, and the corresponding comment in the full demo.

One slight note: assigning to the video planes means creating an empty VideoFrame first.  I've found that using from_numpy_buffer is faster, since it not only avoids a blit (3 ms for a 4k screenshot), but it also avoids creating empty planes when you create the empty VideoFrame.

demos/video-capture-simple.py
jholveck: It does indeed work either way.  Is there an advantage you see to muxing the packets one at a time?

demos/video-capture-simple.py
jholveck: I'll add a note about the terminology to that comment.

Again, passing a list works fine.

demos/video-capture-simple.py
jholveck: We do support them as command line flags on the full demo.

I didn't realize that "libx264" wasn't a valid codec on the Windows build of libavcodec.  Does it go to the Windows Media Foundation instead of built-in libraries or something?

jholveck:
> @jholveck should we do checks for third-party modules?

I think that doing explicit checks like that might be more mess than success.  But I can easily add a comment above where we import third-party modules giving the right pip command to install them.

demos/video-capture-simple.py
jholveck: I think this is probably explained well for beginners in the comments at the top of the file, and explained sufficiently fully in the comments at the top of the full demo (the one without -simple).  Do you disagree?

I'll look forward to seeing your comments on the full demo too!

jholveck:
> > Overall, it works great! I see a small difference in colors, on Linux, but I guess this is due to the JPEG compression.
> You might try turning on DISPLAY_IS_SRGB.

By the way, this relates to #207.  I think that tagging screenshots with the display's colorspace would be a useful future addition to MSS, for just this sort of thing.

BoboTiG:
> > @jholveck should we do checks for third-party modules?
> 
> I think that doing explicit checks like that might be more mess than success. But I can easily add a comment above where we import third-party modules giving the right pip command to install them.

Yes, lets keep it simple then: at the top of the file, one line to install everything.

BoboTiG:
Out of curiosity, do you plan to add more stuff in that PR? Wondering if you keep it as a draft for a specific reason :)

jholveck:
> Out of curiosity, do you plan to add more stuff in that PR? Wondering if you keep it as a draft for a specific reason :)

Not specifically, but I've asked my colleague, @bboudaoud-nv , to review this.  He's worked with a number of internal devs who use MSS and other libraries for video encoding, so I wanted to get any insights he might have.  He's already provided comments on the simple demo; I've asked him to also look at the full one too.

Once he's done his review, I expect that I'll be ready for your final review and commit.

Edit: Actually, I'll probably also fix what's needed to make it Windows-compatible, as he mentioned in his review comments.

demos/video-capture-simple.py
jholveck: I just tried the script as-is on a Windows box, and it seemed to work fine.  I also checked `python -m av --codecs` and it lists libx264 in there.  Were you having problems?

BoboTiG:
> > Overall, it works great! I see a small difference in colors, on Linux, but I guess this is due to the JPEG compression.
> 
> You might try turning on DISPLAY_IS_SRGB.

Actually, DISPLAY_IS_SRGB being True or False changes nothing. and that's OK, I was just noting the fact. No need to provide a script to handle all cases.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+# - P-frames: changes from previous frames.
+# - B-frames: changes predicted using both past *and future* frames.
+#
+# B-frames are why "the order frames are encoded/decoded" can differ from "the order frames are shown".  That leads
+# directly to timestamps.
+#
+# Timestamps (PTS/DTS)
+# --------------------
+#
+# Every frame has a *presentation timestamp* (PTS): when the viewer should see it.
```
bboudaoud-nv: Might be worth noting a PTS is an integer value scaled by the time base (and maybe introduce time base before this section for logical flow)

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+# must set them consistently where you generate timestamps.
+#
+# See <https://pyav.basswood-io.com/docs/stable/api/time.html>
+#
+# This demo uses a time base of 1/90000 (a common MPEG-derived choice).
+#
+# Constant Frame Rate (CFR) and Variable Frame Rate (VFR)
+# -------------------------------------------------------
+#
+# Many video files run at a fixed frame rate, like 30 fps.  Each frame is shown at 1/30 sec intervals.  This is called
```
bboudaoud-nv, line 74: Might be good to point out in cases like this the time base can just be the frame rate/time and then the PTS values are simply integers increasing by 1 at each frame

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+    # 1/fps sec each frame, since we want to also account for the time taken to get the screenshot and other overhead.
+    #
+    # Repeatedly adding small floating-point numbers to a total does cause some numeric inaccuracies, but it's small
+    # enough for our purposes.  The program would have to run for three months to accumulate one millisecond of
+    # inaccuracy.
+    next_frame_at = time.monotonic()
+
+    # Keep running this loop until the main thread says we should stop.
+    while not shutdown_requested.is_set():
+        # Wait until we're ready.  This should, ideally, happen every 1/fps second.
```
bboudaoud-nv, line 181: Same as other script I'd be tempted to capture then sleep for the remaining time to retain per-call perf/avoid slowing over the target rate.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+        #
+        # `np.frombuffer()` creates an ndarray *view* of an existing buffer (no copy).  Reshaping also stays as a
+        # view.
+        #
+        # PyAV's `VideoFrame.from_ndarray()` always copies the data into a new frame-owned buffer.  For this demo we
+        # use the undocumented `VideoFrame.from_numpy_buffer()`, which creates a `VideoFrame` that shares memory with
+        # the ndarray.
+        ndarray = np.frombuffer(screenshot.bgra, dtype=np.uint8)
+        ndarray = ndarray.reshape(screenshot.height, screenshot.width, 4)
+        frame = av.VideoFrame.from_numpy_buffer(ndarray, format="bgra")
```
bboudaoud-nv, line 244: I'm not 100% sure what the type of `screenshot.bgra` is but you can do direct plane assignment to speed this up a bit/avoid the `av.VideoFrame.from_[x]` call.

This is an option, a more performant model might be something like:
```python
frame = av.VideoFrame(format="bgr24")
frame.width = screenshot.width
frame.height = screenshot.height
frame.planes[0] = screenshot.bgra[:,:,0] (or similar)
frame.planes[1] = screenshot.bgra[:,:,1]
...
```
If you can do this you avoid the conversion to/back from the np array and speed up encode slightly

This doesn't allow things like reformatting or using the numpy conversion, but is the fastest model I've seen up to now

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+
+        # Set the PTS and time base for the frame.
+        if first_frame_at is None:
+            first_frame_at = timestamp
+        frame.pts = int((timestamp - first_frame_at) / TIME_BASE)
+        frame.time_base = TIME_BASE
```
bboudaoud-nv, line 263: I don't believe you need to set the individual frame time bases here, just the PTS value. In fact I think you may even be able to set the `frame.time` and let it handle the conversion to PTS here, but I haven't tried that lately.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+        # etc.
+        with av.open(filename, "w") as avmux:
+            # We could initialize video_stream in video_encode, but doing it here means that we can open it before
+            # starting the capture thread, which avoids a warmup frame (one that takes longer to encode because the
+            # encoder is just starting).
+            #
+            # The rate= parameter here is just the nominal frame rate: some tools (like file browsers) might display
+            # this as the frame rate.  But we actually control timing via the pts and time_base values on the frames
+            # themselves.
+            video_stream = avmux.add_stream(codec, rate=fps, options=CODEC_OPTIONS)
```
bboudaoud-nv, line 452: The use of the `rate` arg here with the fps may in some cases set a different time base/assume a CFR encoding. I tend to avoid `rate` here as its not required.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+            # If the encoder accepts BGRA input (e.g., h264_nvenc), we can hand it MSS's BGRA frames directly and
+            # avoid an extra pre-conversion step on our side.
+            #
+            # If the encoder doesn't accept BGRA input (e.g., libx264), PyAV will insert a conversion step
+            # automatically.  In that case, we let the codec choose the pix_fmt it's going to expect.
+            #
+            # Note: the alpha channel is ignored by H.264.  We may effectively be sending BGRx/BGR0.  But PyAV's
+            # VideoFrame only exposes "bgra" as the closest supported format.
+            if any(f.name == "bgra" for f in video_stream.codec.video_formats):
+                video_stream.pix_fmt = "bgra"
```
bboudaoud-nv, line 502: I'm not sure how the "a" is supported, for some reason I thought the format I've used is `bgr24` but perhaps not. Comment does a good job of capturing this.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+            stage_mux = PipelineStage(
+                name="stream_mux",
+                in_mailbox=mailbox_packet_to_mux,
+                target=partial(mux, avmux),
+            )
+
+            stage_mux.start()
+            stage_show_stats.start()
+            stage_video_process.start()
+            stage_video_encode.start()
```
bboudaoud-nv, line 553: In theory these could probably be a single thread. You do get some speed up here, but I'd guess `encode` is the cycle hog and the rest are fairly fast (`process` may have a bit less work to do if you directly access video frame planes). Definitely not required/more for readability than perf (this is better perf)

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+def video_encode(
+    video_stream: av.video.stream.VideoStream, frames: Iterable[av.VideoFrame]
+) -> Generator[Sequence[av.Packet], None, None]:
+    for frame in frames:
+        yield video_stream.encode(frame)
+    # Our input has run out.  Flush the frames that the encoder still is holding internally (such as to compute
+    # B-frames).
+    yield video_stream.encode(None)
```
bboudaoud-nv, line 281: You want to be careful w/ flushing the stream here. If you run out of frames IIRC you want to be careful w/ calling this flush midstream. It may write some sort of end of file signal to the video. May want to not wait for no frames here, but a shutdown signal?

demos/video-capture.py
jholveck: I don't quite get that concept.  Can you elaborate on the difference?

One reason for doing the sleep right before the capture is because the capture is what we want to do at precise intervals.  Putting the sleep here means that jitter in the time taken by other steps (such as blocking in the yield for the mailbox to become empty) doesn't translate into the capture interval.

I'm still not clear on why you are saying that this would impact per-call perf and slow the overall rate.  If we were sleeping 1/30 sec each time, I'd agree with you, but here we're sleeping until 1/30 sec since the previous frame's target time.  Is there something I'm missing?

demos/video-capture.py
jholveck: Can you elaborate on those circumstances?  I haven't seen that happen when I also provide an explicit time base on the stream, as long as I assign the time base after setting the rate.

As the comment mentions, the rate= parameter does set a nominal frame rate for the file, which some tools use for quick display or estimation, or to convert to CFR.  It seems to make sense to include that in the metadata if possible.  But if it causes problems, then yeah, I'll eliminate it.

demos/video-capture.py
jholveck: If we run out of frames, then by definition, we're not midstream.  We're at the end of the video stream.  The `for` loop doesn't exit until the source exits and closes the mailbox.  In other words, the source's `StopIteration` is acting as a shutdown signal.

IOW: if the next frame isn't ready at the top of the `for` loop, then the input iterator will block, not terminate.

demos/video-capture-simple.py
bboudaoud-nv: I haven't tried `libx264` for Windows, but if it works I'm good with it!

demos/video-capture-simple.py
bboudaoud-nv: I think this is just what pyav recommends in [their examples](https://pyav.org/docs/stable/cookbook/numpy.html#generating-video)

demos/video-capture-simple.py
bboudaoud-nv: Ah I didn't realize the `from_numpy_buffer()` was faster, if this is the case I'm mistaken and we should stick with what you have. I find direct plane assignment to be a bit annoying/poorly documented

demos/video-capture-simple.py
bboudaoud-nv: I believe that currently we sleep right up until the next frame, if there is some unknown amount of time from ending the sleep to capturing the frame then the time the frame is actually captured is subject to some variability, if you first capture the frame, then sleep for the remaining time to target you at least align the start of the capture calls as closely as you can. 

Maybe I'm thinking about how this works wrong? What happens in the case where the requested FPS is higher that the rate MSS can capture at? Wouldn't this capture repeated frames at max rate until it "catches back up"?

demos/video-capture-simple.py
bboudaoud-nv: I think having this in the simple demo makes sense now that I've reviewed the more complex one. I now see you've done a nice job explaining this/avoiding the time base here. One thing to think about would be the edge case of capture not being able to keep up with the requested frame rate?

demos/video-capture.py
jholveck: To answer your question about the type, it's `bytes`.  (I'm looking at making it a `memoryview` in the future to avoid the blit.)  It's just a packed linear array of BGRA bytes.  We do also have other ways to access the data: for instance, using the NumPy-compatible `__array_interface__` will give a HWC shape (without a copy; it's just giving that as the array shape specification).

`frame.planes` is a tuple; you can't assign to its elements.  You also can't assign to `frame.planes` either.  In fact, planes is a `@property` that will create a new tuple of new `VideoPlane` objects each time you access it.  (I've checked in PyAV 15.1.0 and 16.1.0, and a quick look at the git history seems to confirm that it's been read-only for over seven years.)

Additionally, `bgr24` is a packed, not planar, format.  It expects a single plane of BGR packed pixels, not three planes.  `gbrp` is the usual planar RGB format, although planar formats are much more common for YUV.
```
In [15]: av.VideoFormat("bgr24").is_planar
Out[15]: False

In [16]: frame = av.VideoFrame.from_image(Image.new(mode="RGB", size=(16, 16), color="red"))

In [17]: frame.format
Out[17]: <av.VideoFormat rgb24, 16x16>

In [18]: frame.planes
Out[18]: (<av.VideoPlane 768 bytes; buffer_ptr=0x279a000; at 0x7fb7c6c08450>,)

In [19]: list(memoryview(frame.planes[0])[:10])
Out[19]: [255, 0, 0, 255, 0, 0, 255, 0, 0, 255]
```

I do have vague memories of creating an empty `VideoFrame` and populating the plane, but I can't find those experiments (I probably deleted them at some point), and I can't remember the specifics of how I did it; possibly using `copy_bytes_to_plane`.  I did find, to my surprise at the time, that it was slower than the alternatives.

The main reason I'm using NumPy as an intermediary is to avoid copying the data at all.  While `from_ndarray` will blit the data, the undocumented `from_numpy_buffer` will reuse pointers into the ndarray.  I haven't found any other way to avoid copying the pixel data when using PyAV.  A 4k screenshot is a 32 MB buffer, and that takes about 3.12 ms to just memcpy that much data (on my Ryzen 7 2700X with DDR4-2133).  Additionally, even with non-temporal stores (which Python won't be using for a list slicing operation), it's still going to blow away even the L3 cache (or whatever your LLC is), and hog the DRAM bandwidth from all the other cores.  When using `from_numpy_buffer`, it can become just a pointer copy.

I did verify that using `from_numpy_buffer` is faster than even using `VideoFrame.from_bytes` directly (which we can't actually use because it only supports RGBA - not BGRA - and Bayer formats, but it's the lowest-overhead mechanism other than from_numpy_buffer).

The easiest way to test this is in `video-capture-simple.py`: since that's single-threaded, the time spent creating the frame shows up very directly in the FPS numbers.  It's easy enough to test different ways of creating the `VideoFrame` there.

This can really be a significant part of the time, and this is how we'll be recommending end users use MSS, so it's worth making sure that we've got it right.  We want our end users to get the best performance they can.

Do you have the code that you're thinking of, to verify the mechanism?

demos/video-capture.py
jholveck: It's kinda weird.

The `h264_nvenc` codec advertises that it accepts several RGB-based formats, including `bgr0` and `bgra`.  (Notably, it doesn't accept `bgr24`, since it really wants word-aligned pixels.)  You can get the complete list with `ffmpeg -h encoder=h264_nvenc`; it's near the top of the output.  I'm guessing it includes `bgra` in there because a lot of times there will be alpha-channel data in the input which, like all H.264 encoders, h264_nvenc will just ignore.

If the format you specify in `pix_fmt` is different than the format of the frame, then PyAV will implicitly insert a CPU-side reformatter in the pipeline.  That reformatter runs on the CPU, and can slow down the pipeline significantly.

That's why I keep everything in bgr0/bgra instead of bgr24: for libx264, it'll have to insert the reformatter either way (libx264 doesn't accept any RGB formats, just YUV).  For h264_nvenc, we want to give it frames in an acceptable RGB format, so it will do the YUV conversion on CUDA cores, instead of on the CPU.

`VideoFrame.from_numpy_buffer` only allows certain formats.  One of these is `bgra`.  It does not, however, accept `bgr0` (which I'd argue is a bug in PyAV).  If we created frames in `bgra` format and set the video stream's `pix_fmt` to `bgr0`, then PyAV would do a CPU-side conversion of the image.

Since the alpha channel is ignored, then it's ok to say they frames are actually BGRA when they're really BGR0.  It may seem kinda goofy, but that's the compromise we have to both use `from_numpy_buffer` (for zero-copy throughput), and avoid PyAV inserting a CPU converter.

You may well have been using `bgr24`.  You could have also been using `yuv420p`, or `nv12`, or whatever.  PyAV will still do something that is correct - convert your frames from whatever format they're in to bgr24 or yuv420p or whatever before sending them to the encoder - but slower under h264_nvenc, since it includes the additional CPU-side conversion step.

demos/video-capture.py
jholveck: Yes, `encode` is definitely the bottleneck.  But the performance difference multithreading makes can be significant, in the range of 10-25%.

Capturing a 4k screen on my home computer while idle:
- libx264, single-threaded: 13.5 fps
- libx264, multi-threaded: 15.3 fps (13% improvement over single-threaded)
- h264_nvenc, single-threaded: 35.6 fps
- h264_nvenc, multi-threaded: 43.5 fps (22% improvement over single-threaded)

The `video-capture-simple.py` demo is single-threaded, for readability.  The `video-capture.py` demo is multi-threaded, to show best practices for serious applications.

Edit: The reason I separate encode from the others is because it's the long pole.  The reason I don't combine capture and process is to have one thread, exclusively used for capture, to help minimize the timing variance.  The reason I have a separate muxing stage is because an earlier version of the code also had an audio stream also feeding the same mux stage, and it seems like something end users might want to add themselves.  (I don't have an audio stream in this demo because audio capture is quite platform-specific.)  But really, there are certainly many ways to structure the pipeline that would work equally well.

I've thought about adding some code to record the percentage of wall time that a thread is processing, blocked on send, and blocked on receive.  That would give a bit more information about structuring things.  But right now, the CPU% is a reasonable proxy for that, so I just print the thread IDs so you can run `top -H` and see.

demos/video-capture-simple.py
jholveck: The capture is rarely the problem in the current versions of MSS; encoding is.  But your point is taken: what happens when the whole system can't keep up?

In the simple demo, it just ends up creating a sped-up Benny Hill-like video: when it's CFR, there's not much else you can do.  The comments at the top of the file explain this.

In the full demo, it's making a VFR video, and tagging the frames with the correct PTS.  So even though it's not able to keep up with the requested frame rate, it's still giving you a best-effort video.

Is there anything else you think needs addressing for this case?

demos/video-capture-simple.py
jholveck: > I believe that currently we sleep right up until the next frame, if there is some unknown amount of time from ending the sleep to capturing the frame then the time the frame is actually captured is subject to some variability, if you first capture the frame, then sleep for the remaining time to target you at least align the start of the capture calls as closely as you can.

I think that's backwards, if I'm understanding you right.  If you put the sleep right after the capture, then the variability in the time for the encode is going to be propagated into the timing of the captured frames.  The idea of putting the sleep before the capture is to have the sleep absorb the variability from all the other parts: if the encode takes 2 ms longer than usual, then the sleep will be 2 ms shorter than usual, and everything stays on schedule.  If we put the sleep after the capture, then that would mean that an encode taking 2 ms longer than usual would also make the next capture be 2 ms later than usual.

To put it differently, the sleep is the step that can absorb any time variation in the whole loop.  It's the "time cushion": as much as possible, it will absorb any variation in the loop's duration.  It's what makes all these rows, which represent loop iterations, all the same width (cycle duration):
```
|-capture--|-encode----------|-mux-|-sleep------|
|-capture---|-encode------------|-mux-|-sleep---|
|-capture-|-encode-----------|-mux--|-sleep-----|
```
(Here, I put the sleep at the end of the cycle for clarity, but it's a cycle, so it's all the same.)

The sleep is waiting until the right phase of the 30 Hz cycle to exit.  It's aligning its exit times to that point; its exit times will (if the system can keep up) all be at 30 Hz intervals.  That's why each loop is the same duration.

The capture is the thing that is time-critical.  That's what we want to make happen at exactly the right time.  That's why we put the sleep right before the capture: the sleep has flexibility in how long it's sleeping, so can be used to align the start of the capture to 1/30 fps intervals.

Let's suppose we took those same example times and put the sleep right after the capture:
```
|-encode----------|-mux-|-capture--|-sleep------|
|-encode------------|-mux-|-capture---|-sleep---|
|-encode--------|-mux--|-capture-|-sleep--------|
```
(Again, I rotated the cycle for clarity.)

Here, the sleep is still exiting at 30 Hz intervals.  Each cycle is still taking 1/30 sec.  But now, the start times of the capture are misaligned.  There's jitter in when we invoke the capture.  That's jitter that was introduced mostly by the encoder in this example.  The sleep is still absorbing all the timing jitter, to keep a constant 30 Hz cycle, but it only can do so after the capture's start time has been impacted by the jitter.

> Maybe I'm thinking about how this works wrong? What happens in the case where the requested FPS is higher that the rate MSS can capture at? Wouldn't this capture repeated frames at max rate until it "catches back up"?

The encode, not the capture, is the bottleneck.  But to answer your question, yes, that's exactly what this simple demo will do: if there's a heavy spike in load, then once things recover, it will run at max rate to catch back up.  If possible, the effect is to make the overall file's mean frame rate as close to the requested average as possible.

This is a bit naive, of course.  You typically don't want the behavior you're talking about.  Ideally, you'd absorb slight jitter to average things out, but not to the point of having big bursts to catch up.

That's why the full demo caps this.  If more than one frame's worth of backlog is accumulated, then it wipes that out (preserving phase to keep the frame cadence smooth).  It won't do a long burst of backlogged frames.  But it will still try to smooth out small amounts of jitter, and keep the mean frame rate as close to desired as possible.

The comment just above these lines in the source does point out that the algorithm in the full demo is better, and what you just described in your question is exactly why.

demos/video-capture-simple.py
jholveck: I would recommend against using that site: it's for PyAV 9, not 16.  See also https://github.com/PyAV-Org/PyAV/issues/1574 for more on that.  The official docs are at https://pyav.basswood-io.com/docs/stable/ .

[This example](https://pyav.basswood-io.com/docs/stable/cookbook/basics.html#recording-the-screen) takes the other approach (although it incorrectly refers to the list as `packet`, singular).  The main reason I'm passing it a list is because [the docs for mux](https://pyav.basswood-io.com/docs/stable/api/container.html#av.container.OutputContainer.mux) refer to it in the plural, and specifically have a separate `mux_one` method.

Although looking at the source code, it's clear that there's been some transition issues around that.  Indeed, `mux` just does a for loop calling `mux_one` if you pass it a list, but was clearly designed so that you could pass the output of `encode` straight to `mux`.

demos/video-capture-simple.py
bboudaoud-nv: I think for the simple demo this is just fine behavior. No additional comments!

demos/video-capture-simple.py
bboudaoud-nv: Yeah this is a difference in desired outcome for the video artifact. In the backup case this approach (incrementing the target time) ends up producing "maximum Benny Hill effect" until it catches back up. Not a problem, but timing each iteration of the loop itself will at least aim for the individual frame times to stay consistent. Again, totally fine with this as an example, but may want to be clear about the behavior on systems that don't perform well

demos/video-capture-simple.py
bboudaoud-nv: Yeah that's an interesting demo, they are treating screen capture like a remuxing problem (not sure if platforms other than mac support this), which while do-able doesn't actually need to encode frames. The corresponding example from the new docs for generating video is [here](https://pyav.basswood-io.com/docs/stable/cookbook/numpy.html#generating-video) I don't believe they changed any code.

demos/video-capture.py
bboudaoud-nv: This is the same idea as above, the goal for video is to have each frame be "captured" 1 / fps apart. The idea that you fall behind 1-off then get many short frames is less desirable than simply resuming regularly spaced frames after. As you point out technically which side of the capture this is on isn't all that important. But if you want to maintain frame rate as best as possible I'd do something like:
```python
dt = 1 / fps
while not shutdown_requested.is_set():
  now= time.time()
  screenshot = sct.grab()
  dur = time.time() - tstart
  yield screenshot, now
  remaining = dt - dur
  if remaining > 0:
    time.sleep(remaining)
```
This way each frame is timed independently and one long frame doesn't cause impacts on other frames. (e.,g., the fps will resume/not "catch up" on the other side of the hitch)

bboudaoud-nv: One thing I'm assuming here is that there is more runtime variation in the `sct.grab()` call duration than there is in the `sleep()` call here. That may not be true on all platforms.

demos/video-capture.py
```diff
@@ -0,0 +1,588 @@
...
+    # Repeatedly adding small floating-point numbers to a total does cause some numeric inaccuracies, but it's small
+    # enough for our purposes.  The program would have to run for three months to accumulate one millisecond of
+    # inaccuracy.
+    next_frame_at = time.monotonic()
+
+    # Keep running this loop until the main thread says we should stop.
+    while not shutdown_requested.is_set():
+        # Wait until we're ready.  This should, ideally, happen every 1/fps second.
+        while (now := time.monotonic()) < next_frame_at:
+            time.sleep(next_frame_at - now)
```
bboudaoud-nv, line 183: This seems biased towards likely to oversleep the goal. Not a problem for relaxed CFR video, but certainly something to consider. On many platforms threaded sleep is a promise to sleep for "at least [X]" but not exactly/under [X]. For example if you do sleep to within say < 1ms of the goal you might just wanna call that good enough here.

demos/video-capture.py
bboudaoud-nv: The exact details aren't hanging around too well, but I recall something like:
```python
# This is a 4 channel bgra image as uint8
image: ndarray 
# Create the frame and update its planes
frame = VideoFrame(width=w, height=h, format="bgra")
frame.planes[0].update(image.tobytes())
```

demos/video-capture.py
bboudaoud-nv: I think this is more confusing to folks than actually a problem. I believe what happens here is the time base gets set using `rate` then this just gets overridden later in the process and PTS values are assigned to each frame. I'm not sure if adding `rate` to this constructor has any other effect to the encoded video.

demos/video-capture.py
bboudaoud-nv: This makes sense to me, looking at previous work I've done with `bgra` video encode from the windows-capture package, this is the value I used

demos/video-capture.py
bboudaoud-nv: Ah okay I think I'm not totally up on how these threads are coordinated then, I guess this gets triggered by the end of the process call here and the in mailbox cannot be empty during normal operation?

bboudaoud-nv:
I'm definitely picking nits at this point, this code is very well written and should be a nice example of how to (actually) screen capture video with pyav as opposed to the weird muxing example they provide in their docs!

demos/video-capture.py
jholveck: It can be empty, but in that case the input iterator will block until something does come available.  The comments at the top of `common/pipeline.py` explain that part.

It's similar to doing something like `for line in input_file:` when input_file is a socket or pipe or something.  If there's presently no data (but more is presumably on the way), it'll just sit there and wait for more data to arrive.  It doesn't end the iteration until the sender signals that no more is forthcoming.

demos/video-capture.py
jholveck: Yeah, I thought about your precision sleep library when I wrote that.  I decided it wasn’t worth the added complexity for this purpose.  But to speak to your point, I'm afraid I'll need to get a little pedantic.

I believe we've already accounted for *bias*, although not *variance*.  This loop structure is designed to eliminate bias (since we only care about the capture frequency, not phase).

Suppose the `sleep` implementation (just as an example) tends to oversleep by a mean of 2 ms, with a standard deviation of 0 ms: it’s all bias, no variance.

The first loop iteration, let’s say, starts at midnight.  Since `next_frame_at` increments by 1/30 sec each frame, regardless of how long anything in the loop took (including the sleep itself), it will schedule frames at 0 s, 0.033 s, 0.067 s, and 0.100 s.

Now, let’s look at when the sleeps exit, and the captures happen.  The `sleep` implementation doesn’t return until 2 ms after it’s scheduled.  That means the captures will be at 0.002 s, 0.035 s, 0.069 s, and 0.102 s.

These are still at 1/30 sec intervals.  They’re not the exact times that we sent to `sleep`, but they’re still at the desired frequency, just at a different phase.

Now, let’s add in variance.  By definition, the variance term averages to 0.  For now, let’s assume that the sleep still has a mean of 2 ms, but now its variance is always ±1 ms (easier to write the numbers than σ²=1 ms).  It can overshoot by 1 or 3 ms.  So what happens then?

We still are using the same sequence of deadlines, of course: 0 s, 0.033 s, 0.067 s, 0.100 s.  Now, the sleep exits at 0.003 s, 0.034 s, 0.070 s, and 0.101 s.

Our frames are still scheduled at a mean frequency of exactly 1/30 sec, although now with a frequency error of 0 ± 1 ms: jitter.  Unfortunately, that’s not something that we can get rid of without a precision sleep implementation like yours.  However, that timing noise is a lot less of a problem than a systematic error, since that would cause issues with A/V sync, seeking, etc.

I believe the loop structure — incrementing the target time by 1/30 sec each iteration — eliminates the systematic error.  There’s no bias in the frequency.  This is different than the most common naïve implementations I’ve seen: either adding 1/30 sec to the previous loop’s actual start time (`now` in our code), or worse yet, sleeping 1/30 sec each loop.  Both of those will show a bias in the frame timings.

If this code does still allow a bias, a systemic error that has a non-zero mean, then I’d want to look for ways to eliminate that.  But I think the present implementation, as it stands, does eliminate the bias, and ensures that any timing errors in `sleep` only get propagated to the frame timings with their variance, not bias.

jholveck:
> I'm definitely picking nits at this point, this code is very well written and should be a nice example of how to (actually) screen capture video with pyav as opposed to the weird muxing example they provide in their docs!

I'm so glad to hear it!  Thank you for your diligent review!  At this point, I think I'll review your comments again and see if there are ways I can improve the comments to explain some of the sticking points, and then @BoboTiG can merge it.

demos/video-capture.py
jholveck: Ah, gotcha.  Yeah, that incurs a memcpy, which is what I was trying to avoid.  I'll test to be sure, since this is an important thing to accurately present to end users.  One of the few things in this whole demo that's actually directly related to MSS is showing the best way to transfer pixel data from MSS to PyAV and other libraries!

demos/video-capture.py
bboudaoud-nv: I believe `naptime` (our accurate sleep library) should be publicly available shortly. So if you want to use it, it's a low dependency/small library to improve this. It will increase system load marginally w/ more busy wait, but is useful in a case like this.

Generally speaking on Linux `time.sleep()` time variance is lower as the scheduler does a better job of servicing tasks at a fine time granularity. But on Windows as system load increases I've seen > 4 ms sleep "increments" and minimum sleep times of up to 16 ms. This can be a pain when trying to accurately wait for just a few ms.

jholveck: Glad to hear it!

Another bit of good news: in Python 3.11 and later, on Windows 10 and newer, Python will use a high-resolution timer.  This provides a resolution of 100 ns.  I'm not sure how this interacts with the scheduler, but it probably won't be nearly as bad as what you're seeing presently.  I hadn't realized that the Windows sleep implementation was that coarse!

demos/video-capture.py
jholveck: It's hard for me to tell quite when it's used, but it's in places other than the time base.

For instance, it's definitely written to the Matroska (MKV) headers, if you're using that format, hence my comment about it setting a nominal frame rate that "might" be displayed.  (Try `./video-capture.py --fps 123 --output foo.mkv --duration 10 && ffprobe foo.mkv`, and note that "123" is in the nominal frame rate display shown there.)  I think it might be used in an MP4's tmcd atom, but I can't really tell how.  It's like a lot of things in ffmpeg: it's impossible to tell quite what the effect is in all the different use cases, because there are so many different data paths for all these different multimedia formats.

The ffmpeg docs say on this variable (`AVStream::avg_frame_rate`), "May be set by the caller before avformat_write_header()".  So that's why I'm including it.

demos/video-capture.py
jholveck: You're absolutely correct that you can set just the PTS value.  If the frame doesn't have a time_base set, then PyAV will assume the destination stream's time_base when it encodes the frame.  (This is a feature of PyAV, not ffmpeg, as far as I can tell.)

However, you do need to set the time base if you want to use `frame.time`, which can be quite handy for debugging.  It's also necessary if you want to, for instance, put your frame times in nanoseconds (for simplicity) and just have them get converted to the stream time_base at encoding time; that's what I've done in some code I've written.

It's not mandatory, but it does seem to be a reasonable practice, rather than relying on the default "whatever I get put into" behavior.  Thanks for the point, but I think I'll leave it as-is.  I've gotten burned in the past by not setting the time_base in all the right places, so I think implicit is reasonable here.

FWIW, you can't set `frame.time`.

jholveck:
@BoboTiG I'm happy with merging it as it stands.  I think @bboudaoud-nv and I have some different ideas about how to handle the finer details of the timing, but nothing that's significant enough to worry about for a demo.  I may come back and change the relevant code later, but as it stands I think it's in great shape to commit.  Then we can get back to focusing on improvements to functionality.

PR #457, opened 2026-01-17
Fix tests so they work with all multi-monitor setups

halldorfannar:
The setup of having a monitor to the left or above the primary monitor was not working correctly.  A number of tests would fail.

This fixes 25 test failures I would otherwise get on my multi-monitor Windows machine.  These changes are internally facing and therefore do not warrant a CHANGELOG.md entry or documentation update.

- [x] Tests added/updated
- [x] `./check.sh` passed

### DIFF
--- a/src/mss/base.py
+++ b/src/mss/base.py
@@ -186,9 +186,6 @@ class MSSBase(metaclass=ABCMeta):
                 "height": monitor[3] - monitor[1],
             }
 
-        if monitor["left"] < 0 or monitor["top"] < 0:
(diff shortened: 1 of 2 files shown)


jholveck:
Sorry about that!  I recently added that runtime check, not realizing that Windows would assign monitors to the left of the primary monitor negative coordinates.  Thanks for catching it!

BoboTiG:
Thank you @halldorfannar, that would have been an bad regression if I had cut a release with the code you fixed!

PR #459, opened 2026-01-18
Add `check.ps1` for Windows

### LINKED ISSUE #458: An equivalent of `check.sh` for Windows
The developer guidelines ask for us to run `./check.sh` before creating a PR.  This is a good guideline.  But on Windows it is not expected that developers run `.sh` scripts (it's possible though, via Git Bash shell, but this will feel unfamiliar and some may not even know about this possibility).

I suggest we make a change so that this works well on Windows too.  There are a few ways to do this, and I therefore haven't created a PR yet.  There is no right or wrong answer, so I feel the maintainer should make the decision 😄 

1. Create a `check.ps1` script (or `check.bat` if preferred) that contains the same execution steps as `check.sh`
2. Create a `check.py` script that replaces `check.sh`. The developer guideline would be changed to run `python check.py` - the benefit of this approach is that we only have one file to maintain but running the check step is now slightly less ergonomic.
3. A fusion of the above, create a `check.ps1` (or `check.bat`) that just contains `python check.py` internally (this way we would still have two files that are native for each operating system but the code would all be maintained in the one `check.py` file).

I have no horse in this race!  Please pick the winner.

halldorfannar:

This makes it easy for developers on Windows to run the required checks.

Changes are internally facing so CHANGELOG.md was not updated.

Fixes #458

- [x] Documentation updated
- [x] `./check.sh` passed

### DIFF
--- a/src/mss/base.py
+++ b/src/mss/base.py
@@ -186,9 +186,6 @@ class MSSBase(metaclass=ABCMeta):
                 "height": monitor[3] - monitor[1],
             }
 
-        if monitor["left"] < 0 or monitor["top"] < 0:
-            msg = f"Region has negative coordinates: {monitor!r}"
-            raise ScreenShotError(msg)
         if monitor["width"] <= 0 or monitor["height"] <= 0:
             msg = f"Region has zero or negative size: {monitor!r}"
             raise ScreenShotError(msg)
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks @halldorfannar !

PR #465, opened 2026-01-22
New demo: cat detector

jholveck:
This will detect if a cat is on the screen.  By which I mean displayed on the screen, not sitting on your laptop.

This is meant as a simple demo of using MSS for AI.  It works as-is, but needs to be documented, and there's some bits that could do with cleanup.

There are a lot of additional features that could be added, such as showing a window with bounding boxes, but that's probably more complexity than is called for here.

- [ ] Tests added/updated - N/A
- [X] Documentation updated
- [X] Changelog entry added
- [ ] `./check.sh` passed - N/A (doesn't check demos)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ See Git commit messages for full history.
 - Windows: improve error checking and messages for Win32 API calls (#448)
 - Mac: fix memory leak (#450, #453)
 - improve multithreading: allow multiple threads to use the same MSS object, allow multiple MSS objects to concurrently take screenshots, and document multithreading guarantees (#446, #452)
-- Add full demos for different ways to use MSS (#444, #456)
+- Add full demos for different ways to use MSS (#444, #456, #465)
 - :heart: contributors: @jholveck, @halldorfannar
 
 ## 10.1.0 (2025-08-16)
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -225,3 +225,4 @@ scenarios.
 These include:
 - MP4 video capture with encoding using PyAV (FFmpeg bindings)
 - Live streaming to a TinyTV as MJPEG
+- Detect images of cats on the screen


BoboTiG:
I like it, great inspiration!

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+# computer vision.  These will look at a picture that has a single foreground object, front and center, and try to
+# identify what type of object this is: a cat, person, bicycle, etc.
+#
+# An object detector looks at an image and identifies _multiple objects_ within it.  Instead of assigning a single
+# label to the whole image, saying "this is a picture of a cat", it might say "there is a cat here, and a bicycle over
+# there," and provide some basic information about each one.  This is, for instance, what a self-driving car uses to
+# identify what it's seeing on its cameras.
+#
+# For this demo, we want to tell if a cat is anywhere on the screen, not if the whole screen is a picture of a cat.
+# That means that we want to use an detector, not a classifier.
```
halldorfannar:
```suggestion
# That means that we want to use a detector, not a classifier.
```

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+#
+# Finally, *COCO_V1* indicates that the model was trained on the COCO dataset, a widely used community benchmark for
+# object detection.  COCO contains hundreds of thousands of labeled images covering 80 common object categories (such
+# as people, animals, and vehicles), along with a small number of additional placeholder categories that appear as
+# "N/A" in the model metadata.
+#
+# Performance
+# ===========
+#
+# The biggest determinant of performance is whether the model runs on a GPU or on the CPU.  GPUs are extremely
```
halldorfannar: Maybe we should mention here, right away, that this particular model will work on both?  I know this becomes clearer in the end of this section, when GPU vs CPU performance comparisons are discussed.

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+# time regardless of the original screen resolution.
+#
+# With a CUDA-capable GPU, this demo's main loop typically runs in around 100 ms per frame (about 10 fps).  When run
+# on the CPU, the same work takes roughly 5000 ms per frame (about 0.2 fps).
+#
+# Cached Data
+# ===========
+#
+# The first time you run this demo, Torchvision will download a 167 MByte DNN.  This is cached in
+# ~/.cache/torch/hub/checkpoints on Unix.  I'm not sure where it's cached on other platforms, but it will tell you.
```
halldorfannar: Third person was used earlier in the text.  I would avoid switching to first person here.  Perhaps just go with this:
```suggestion
# ~/.cache/torch/hub/checkpoints on Unix.  If you want to know where the cache is stored on other platforms, this information will be displayed after downloading the DNN.
```

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+import time
+
+# You'll need to install PyTorch and TorchVision, and the best way to do that can vary depending on your system.
+# Often, "pip install torch torchvision" will be sufficient, but you can get specific instructions at
+# <https://pytorch.org/get-started/locally/>.
+import torch
+import torchvision.models.detection
+import torchvision.transforms.v2
+
+# You'll also need to "pip install mss pillow".
```
halldorfannar: Rather than assume the user is leveraging `pip` (I'm a fan of `uv`) I would suggest the more general:
```suggestion
# You'll also need to install mss and pillow.
```
This also aligns with the earlier text where `pip` is suggested but it is left open for the user how to do this specifically.

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+# The model will identify objects even if they only vaguely look like something.  It also tell us a score of how
+# certain it is, on a scale from 0 (not a cat) to 1 (very confidently a cat).  To prevent false positives, we set a
+# threshold and ignore any results below it.  The score doesn't have any real external meaning: to pick the cutoff,
+# you just try different images, look at the scores, and get a sense of what seems about right.
+SCORE_THRESH = 0.60
+
+# If an image is too small, then it's got a pretty decent chance of being a false positive: it's hard to tell if a
+# Discord or Slack reaction icon is a cat or something different.  We ignore any results that are too small to be
+# reliable.  Here, this cutoff is 0.1% of the whole monitor (about 1.5 cm square on a 27" monitor, the diameter of a
+# AA battery).  Like the score threshold, this is just something you try and see what the model seems to be able to
```
halldorfannar:
```suggestion
# AA battery).  Like the score threshold, this is just something you try and see what the model is able to
```

demos/cat-detector.py
```diff
@@ -0,0 +1,384 @@
...
+    # For the model we're using, the preprocessing is simply to standardize the representation: it will convert PIL
+    # images to a tensor representation, and convert all images to floating-point 0.0-1.0 instead of integer 0-255.
+    # Some other models do more preprocessing.
+    #
+    # Fortunately, for its pretrained models, Torchvision gives us an easy way to get the correct preprocessing
+    # function.
+    preprocess = weights.transforms()
+
+    # The labels ("what type of object is this") that the model gives us are just integers; for this model, they're
+    # from 0 to 90.  The English words describing them ("cat") are in a list, stored in the weight's metadata.
```
halldorfannar:
```suggestion
    # from 0 to 90.  The English words describing them (like "cat") are in a list, stored in the weight's metadata.
```

demos/cat-detector.py
jholveck: I prefer very specific commands in user examples, but I'll make the "pip" command an example rather than sounding like a requirement.  I had kind of assumed that the user, if using something other than pip, would know how to adapt the command to their needs.

The reason I pointed the user to the PyTorch "Get Started" page wasn't to handle different package managers; that page still only gives commands for pip (plus building from source, and downloading C++ / Java packages).  The main difference for our purposes is that it gives different --index-url flags depending on whether you want a CUDA, ROCm, or CPU-only build.

I'll make a change to soften the suggestion of package manager, but still leave the specific command present.

demos/cat-detector.py
jholveck: Adding to the next commit

PR #472, opened 2026-02-06
Demo video-capture width and height must be divisible by 2

sjoerdsmink:
When the resolution is not an even number, it generates an error:
```
 File "av/codec/context.pyx", line 200, in av.codec.context.CodecContext.open                                                                                                                           
    File "av/codec/context.pyx", line 220, in av.codec.context.CodecContext.open                                                                                                                           
    File "av/error.pyx", line 424, in av.error.err_check                                                                                                                                                   
  av.error.ExternalError: [Errno 542398533] Generic error in an external library: 'avcodec_open2(libx264)'; last error log: [libx264] height not divisible by 2 (1512x949)
```

Fixes #
Width and height must be divisible by 2, otherwise the encoder will fail. Round down to the nearest even number.

- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] Changelog entry added
- [ ] `./check.sh` passed

### DIFF
--- a/demos/video-capture-simple.py
+++ b/demos/video-capture-simple.py
@@ -78,8 +78,8 @@ def main() -> None:
             video_stream = avmux.add_stream(CODEC, rate=FPS, options=CODEC_OPTIONS)
             # Width and height must be divisible by 2, otherwise the encoder will fail.
             # Round down to the nearest even number.
-            video_stream.width = monitor["width"] & ~1
-            video_stream.height = monitor["height"] & ~1
+            video_stream.width = monitor["width"] // 2 * 2
+            video_stream.height = monitor["height"] // 2 * 2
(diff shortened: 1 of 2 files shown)


jholveck:
I would suggest changing the capture resolution too; not sure what the libraries will do if they don't match, but making them match is probably best.

In the non-simple demo, I would also conditionalize this on the codec being libx264: not all codecs have this limitation (I just tested on h264_nvenc).

sjoerdsmink:
Could you then make those changes? Feel free to close this PR and do it in a separate one. I don't feel confident making the changes, as I can't assess the impact. Doing it this way the frames from MSS will be slightly larger, but PyAV's reformatter seemed to handle the crop automatically.

jholveck:
Opened #474 with these changes.

BoboTiG:
Fixed in #474.

PR #477, opened 2026-02-20
Add identifying information about monitors under Linux

### LINKED ISSUE #153: How can I tell which is the primary monitor in the monitors array?

jholveck:

Adds new attributes to the _monitors dicts.  These are all optional; they’re only populated if they can be probed via XRandR.

- `output`: The output name, as used in XRandR.  This is the main identifier used in some monitor configuration tools, like the Linux xrandr command-line tool.  This is only populated under Linux.
- `name`: The human-friendly monitor name, from the EDID block
- `unique_id`: A platform-specific string that tries to identify the monitor as specifically as possible, across reconnections and configuration changes.  On Linux, this is a URL-query-formatted string including the ID, serial, and dates from the EDID block.
- `is_primary`: True if the monitor is marked as “primary”.  This is where things like the taskbar are expected to appear.

This change also adds support for RandR 1.5 Monitors, a user-configurable concept that aligns with the MSS monitor concept.  This separates the physical CRTC concept from the logical concept, used for things like tiled displays.

This change also adds some small developer-visible improvements: it adds support for InternAtom, and also makes XIDs and other X primitive types comparable (preventing one class of bugs).

Fixes #153 (for Linux)
Related to #469 ; this is the Linux side

- [X] Tests added/updated
- [x] Documentation updated - Handled in #469, I think, but I can add it here
- [x] Changelog entry added - Handled in #469, I think, but I can add it here
- [X] `./check.sh` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 See Git commit messages for full history.
 
 ## 10.2.0.dev0 (2026-xx-xx)
+- Add `is_primary`, `name`, and `unique_id` keys to Monitor dicts for primary monitor detection, device names, and stable per-monitor identification (#153)
+- Add `primary_monitor` property to MSS base class for easy access to the primary monitor (#153)
+- Windows: add primary monitor detection using `GetMonitorInfoW` API (#153)
+- Windows: add monitor device name and unique device interface name using `EnumDisplayDevicesW` API (#153)
 - Linux: add primary monitor detection, monitor device name, unique device interface name, and output name using XRandR (#153)
 - Windows: switch from `GetDIBits` to more memory efficient `CreateDIBSection` for `MSS.grab` implementation (#449)
 - Windows: fix gdi32.GetDIBits() failed after a couple of minutes of recording (#268)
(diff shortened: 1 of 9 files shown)


src/mss/linux/base.py
```diff
@@ -152,47 +185,168 @@ def _monitors_impl(self) -> None:
...
+                0,  # long-offset: 0
+                1024,  # long-length: in 4-byte units; 4k is plenty for an EDID
+                0,  # delete: false
+                0,  # pending: false
+            )
+            if edid_prop.type_.value != 0:
+                edid_block = bytes(xcb.randr_get_output_property_data(edid_prop))
+                edid_data = parse_edid(edid_block)
+                if "display_name" in edid_data:
+                    rv["name"] = edid_data["display_name"]
```
BoboTiG:
```suggestion
                if display_name := edid_data.get("display_name"):
                    rv["name"] = display_name
```

BoboTiG: Maybe could we optimize a bit all those double dict accesses?

jholveck: Fixed.  Thanks for the note; I'd forgotten that := was even available by Python 3.9!

BoboTiG:
Thank you @jholveck!

Issue #220, opened 2022-12-20
Unjustified Screenshot error after closing a Tkinter Toplevel (Linux)


kelltom:
General information:

* OS name: _Ubuntu Linux_
* OS version: __22.04__
* OS architecture: _64 bits_
* Resolutions:
    * Monitor 1: _3440x1440_ (issue occurs at all resolutions)
* Python version: _3.10.6_
* MSS version: __6.1.0__ & __7.0.1__

For GNU/Linux users:

* Display server protocol and version, if known: __X11 server__
* Desktop Environment: _ubuntu:GNOME_
* Composite Window Manager name and version: __X.Org v: 1.21.1.3, compositor: gnome-shell v: 42.5__

### Description of the warning/error

--- **The issue (and solution) is described in detail [here](https://stackoverflow.com/questions/74856512/screenshot-error-xdefaultrootwindow-failed-after-closing-a-tkinter-toplevel/74858212#74858212)** ---

In summary, the following error occurs when trying to instantiate an `mss` object after opening/closing a Tkinter Toplevel window. This error does not occur on Windows.

### Full message

```python
  File "/home/my_project/src/utilities/geometry.py", line 64, in screenshot
    with mss.mss() as sct:
  File "/home/my_project/env/lib/python3.10/site-packages/mss/factory.py", line 34, in mss
    return linux.MSS(**kwargs)
  File "/home/my_project/env/lib/python3.10/site-packages/mss/linux.py", line 297, in __init__
    self.root = self.xlib.XDefaultRootWindow(self._get_display(display))
  File "/home/my_project/env/lib/python3.10/site-packages/mss/linux.py", line 184, in validate
    raise ScreenShotError(f"{func.__name__}() failed", details=details)
mss.exception.ScreenShotError: XDefaultRootWindow() failed

Screenshot error: XDefaultRootWindow() failed, {'retval': <mss.linux.LP_XWindowAttributes object at 0x7f31d8d38ac0>, 
    'args': (<mss.linux.LP_Display object at 0x7f31d8d38740>,)}
```

### Other details

The following code can be used to reproduce the error. Simply run this program and use the GUI button to open a new window that allows you to take a screenshot - then _close_ the pop-up window and attempt to repeat the process.

```python
import tkinter as tk

import mss
import mss.tools


def take_screenshot():
   with mss.mss() as sct:
       screen_part = {"top": 370, "left": 1090, "width": 80, "height": 390}
       sct_img = sct.grab(screen_part)
[...]

PR #480, opened 2026-02-24
Add export methods for PIL and popular array libraries to the ScreenShot object

### LINKED ISSUE #478: Add .to_pil, etc methods to ScreenShot
This is based on a conversation started in #476 .

The proposed methods will let us provide users with easy-to-use best-practice ways to get the data to the desired destination.

The CPU could also have methods to send data to PyTorch and TensorFlow, since those can work with both CPU-resident and CUDA-resident data.  For instance, the cat detector demo uses PyTorch, giving it the MSS pixel data from the CPU, and then either running the neural net on the CPU, or moving the data back to the GPU and running the neural net there.

Here's something to consider in the API design: different libraries have different conventions for how to represent the data.  For instance, even though they all use NumPy, OpenCV wants the channels in BGR order, while Matplotlib, scikit-image, and most others want the channels in RGB order.  For PIL, it's not a problem: we just return an Image, and PIL knows its internal layout.  But for things like `.to_numpy`, it would be good to provide the user with the knobs they'll want to interface with their favorite image processing library.  These knobs aren't hard to implement, and handling them when we're preparing the array is more efficient than the user having to reorder it later.

Here’s a cheat sheet I got from ChatGPT about the expectations of popular image frameworks.  I would verify this before relying on it, but this gives us an idea of what users may need.  (For a head start on verification: its unedited response, with links to references, are at <https://chatgpt.com/s/t_6996cd0b606c8191aba9c2cc6e23dc3f>.)

Just to define some terms:
- **Array type:** The array library the image wants to use (NumPy, PyTorch, etc)
- **Layout:** `HWC` means the axes are [height, width, channels], in that order.  `HWC` is what we provide with `.pixels()`, for instance.  But some very popular frameworks want `CHW`.  (This terminology also often has another dimension if you’re processing N images in a batch, so you’ll often see `NHWC`, but that’s not relevant for MSS.)
- **Channel order:** RGB or BGR, typically.
- **dtype:** The data type that the framework wants for its inputs.  This is almost always either `uint8` if the pixels are 0–255 (like MSS uses) or some type of float pixel in 0–1 (very common for neural nets).

Here’s the list:

| Framework     | Array type | Layout | Channels | dtype  |
| ------------- | ---------- | ------ | -------- | ------ |
| OpenCV        | NumPy      | HWC    | BGR      | uint8  |
| scikit-image  | NumPy      | HWC    | RGB      | either |
| Matplotlib    | NumPy      | HWC    | RGB      | either |
| Torchvision   | PyTorch    | CHW    | RGB      | float  |
| Kornia        | PyTorch    | CHW    | RGB      | float  |
| TensorFlow    | TensorFlow | HWC    | RGB      | float  |
| JAX           | jax        | HWC    | RGB      | float  |

These different needs can easily be accommodated with parameters in to_numpy, to_pytorch, and any other similar methods we create to build array-like objects.

jholveck:
Add `mss.ScreenShot.to_pil`, `to_numpy`, `to_torch`, and `to_tensorflow` methods to `mss.ScreenShot`.

This change enhances integration with popular scientific and AI frameworks, simplifying image export, and improving documentation and examples. The most significant changes are the introduction of new export methods for `ScreenShot` objects, especially `.to_pil()` and `.to_numpy()`, across demos and documentation.

This gives users a set of tested and high-performance methods to easily export the pixel data to their preferred framework for processing.  We can also update these methods as we improve MSS or as the frameworks add capabilities that we can use, allowing users to automatically gain these advantages without changing their code.

The documentation, demos, and examples are all changed to use the new methods.  This simplified a lot of them; it's probably worth revisiting the examples page, possibly removing some of the export examples that are now just one-liners, or expanding on them a little to show usage beyond just the export step.

A new section is added to the usage documentation page, walking through the basic ideas of grabbing a screenshot and exporting the pixel data to another framework.

New tests are added.  The CI/CD GitHub workflow will now install PyTorch and TensorFlow when available to test the relevant methods, but these are not added to the `mss[tests]` optional dependency group.

As an aside: this is the first MSS code change that adds some explicit GPU support.  The new `to_torch` method has a `device=` parameter, and the new `to_tensorflow` method will respect the current TensorFlow device placement policy.  (This asymmetry is because the two frameworks have very different traditions about device placement.)  The change doesn't involve any GPU-specific code, and the initial screenshot's pixel data is still transferred to a CPU buffer, but we now provide a simple way for users to send it back to the GPU for further processing.

Fixes #478 

- [X] Tests added/updated
- [X] Documentation updated
- [x] Changelog entry added
- [X] `./check.sh` passed


BoboTiG:
Python 3.9 will be dropped in the next major release, so no need to take care of test failures.

I think we can do a release without this patch, right?

jholveck:
Certainly!

jholveck:
We can still wait to merge this, but the fix for Python 3.9 was simple, so I went ahead and committed it.

BoboTiG:
You can revert the Python 3.9 fix commit ;)

jholveck:
I've updated this a bit, but should probably review it before committing.  It's been a while!

jholveck:
This is on hold pending a resolution to #536, which will affect how we should document these methods and their sharing semantics.

jholveck:
My last commit added a lot to the docs.  There's a couple more things I want to add to the docs before closing this, even if they're somewhat out of scope for this PR.

BoboTiG:
Great! The documentation is very very good since you both started to work on the project with @halldorfannar 🤟 

Go ahead for out-of-scope commits, I'll merge when you're ready.

jholveck:
Well, that test was weird.  Glad you caught it.

All the stages seem to have completed in seconds.  So I think this was a transient issue in GitHub's orchestration, not a problem with MSS.  Orchestrating CI/CD on macOS is hard.

If it recurs, we can look more at it.  But I'm not concerned yet.

jholveck:
I think this is now feature-complete, and ready for @BoboTiG and @halldorfannar 's reviews.  Sorry it's such a long PR!

demos/tinytv-stream-simple.py
```diff
@@ -97,7 +97,7 @@ def main() -> None:
                     # The next step is to resize the image to fit the TinyTV's screen.  There's a great image
                     # manipulation library called PIL, or Pillow, that can do that.  Let's transfer the raw pixels in
                     # the ScreenShot object into a PIL Image.
-                    original_image = Image.frombuffer("RGB", screenshot.size, screenshot.bgra, "raw", "BGRX", 0, 1)
+                    original_image = screenshot.to_pil("RGB")
```
BoboTiG, line 100: That's so much cleaner that way!

docs/source/usage.rst
```diff
@@ -32,6 +32,131 @@ still available in 11.0, but are also deprecated::
...
+starting from index 1, as well as the full virtual screen (all monitors combined) at index 0.
+
+Once you've decided what you want to capture, you can call :py:meth:`mss.MSS.grab` with the appropriate monitor or
+region.  This will return a :py:class:`mss.ScreenShot` object, which contains the pixel data and other information about
+the screenshot.
+
+For instance, you can capture the first monitor and get a :py:class:`mss.ScreenShot` object like this::
+
+    with MSS() as sct:
+        sct_img = sct.grab(sct.monitors[1])  # Capture the first monitor
```
BoboTiG: We never talk about `sct.primary_monitor`; maybe this line is a good place to start?

src/tests/third_party/array_frameworks/test_numpy_method.py
```diff
@@ -0,0 +1,67 @@
...
+def test_to_numpy_bad_layout() -> None:
+    raw = bytearray([0, 0, 0, 0])
+    shot = ScreenShot.from_size(raw, 1, 1)
+
+    with pytest.raises(ValueError, match='Layout must be "HWC" or "CHW"'):
+        shot.to_numpy(layout="NHWC")  # type: ignore[arg-type]
+
+
+@pytest.mark.parametrize("layout", ["HWC", "CHW"])
+@pytest.mark.parametrize("channels", ["BGRA", "BGR", "RGBA", "RGB"])
```
BoboTiG, line 62: Maybe should we  import `screenshot.Channels` & `screenshot.Layout` and use these values here to never miss a change in the future?

src/tests/third_party/array_frameworks/test_tensorflow_method.py
```diff
@@ -0,0 +1,53 @@
...
+    raw = bytearray([9, 8, 7, 6])
+    shot = ScreenShot.from_size(raw, 1, 1)
+
+    tensor = shot.to_tensorflow(dtype="uint8")
+    assert tensor.dtype == tf.uint8
+    assert (tensor.numpy()[0, 0] == [7, 8, 9]).all()
+
+
+@pytest.mark.parametrize("layout", ["HWC", "CHW"])
+@pytest.mark.parametrize("channels", ["BGRA", "BGR", "RGBA", "RGB"])
```
BoboTiG, line 41: Same here.

src/tests/third_party/array_frameworks/test_torch_method.py
```diff
@@ -0,0 +1,68 @@
...
+    raw = bytearray([5, 6, 7, 8])
+    shot = ScreenShot.from_size(raw, 1, 1)
+
+    tensor = shot.to_torch(dtype=torch.uint8)
+    assert tensor.dtype == torch.uint8
+    assert torch.equal(tensor[:, 0, 0], torch.tensor([7, 6, 5], dtype=torch.uint8))
+
+
+@pytest.mark.parametrize("layout", ["HWC", "CHW"])
+@pytest.mark.parametrize("channels", ["BGRA", "BGR", "RGBA", "RGB"])
```
BoboTiG, line 37: And here.

src/tests/third_party/array_frameworks/test_numpy_method.py
jholveck: Those are types, not lists.  `Literal["A", "B", "C"]` is a type declaration saying that the `Channels` type is one of those three strings.  We could use `typing.get_args()` to list the arguments passed to the Literal type, although that would then break if we later made it a Union or something.  We can do that, but it does make me a little uneasy.

docs/source/usage.rst
jholveck: Great point!  I'm adding that here.

src/tests/third_party/array_frameworks/test_numpy_method.py
BoboTiG: OK, lets forget the idea then.

BoboTiG (approved):

jholveck:
It seems that one of the macOS tests is stuck again.

Also, I previously said the stages had completed, but was mistaken: this is hanging during PyTest.  If the CI/CD logs are complete (that is, not buffering several lines at a time), then it seems like it's now executing `src/tests/test_issue_220.py::test_regression`.

I'm not sure if this is a change in macOS, Python, or what.  It could conceivably be related to my code, but I'm not sure how.

BoboTiG:
This test stressing tkinter, and the OS being macOS, we are not yet ready for all potential suprprises we could hit. Lets see if it happens on a regular basis, for now if retrying manually on timout fixes it, that's good.

jholveck:
The original issue, #220, was Linux-specific.  You also tracked it down to an issue specific to Xlib error handling.  Error handling is one big reason I wanted to migrate to XCB: it has a model that's much easier to work with when you have multiple libraries in a single program, let alone multiple threads.

I'll take a bit of a closer look at the issue and see if maybe it's making assumptions about things like event ordering that tkinter doesn't actually promise, and so has a race condition that just didn't happen to be a problem until recently.  But if I don't find a way to fix the hang, I also wouldn't feel bad about limiting that test to just Linux.

Edit: if I do pursue changes related to this, it'll be in a separate PR.  I just put the timeout in this PR so it didn't keep bugging us as we close it up.

halldorfannar:
Looks good to me!

jholveck:
Ok!  I think there's always more to improve, but I feel like it's at a good point to merge.

PR #503, opened 2026-04-20
Adjust Windows to match the Linux backend setup

halldorfannar:

The Windows implementation now has a backend dispatcher and one implementation, `gdi`, which is the default.

Added a new test to get 100% coverage on the `__init__.py` module for this new Windows package. Also improved one type annotation, since I was already going over all this code.

- [x] Tests added/updated
- [x] Documentation updated
- [x] Changelog entry added
- [x] `./check.sh` passed

### DIFF
--- a/docs/source/release-history/v10.2.0.md
+++ b/docs/source/release-history/v10.2.0.md
@@ -155,7 +155,11 @@ Actual performance improvements will vary depending on factors such as:
 - hardware
 - whether the shared-memory capture path is available
 
-## Windows Capture Improvements
+## Windows Improvements
+
+Windows has received both capture-reliability improvements and a package restructure that prepares the codebase for additional backends.
+
+### Capture with `CreateDIBSection`
 
 The Windows screenshot implementation now uses **`CreateDIBSection`** instead of `GetDIBits`.
 
@@ -165,6 +169,18 @@ Additional improvements include:
 
 - improved Win32 error handling and diagnostics
 - fixes for capture failures during extended recordings
+
+### Package Restructure
+
+`mss.windows` is now a **package** rather than a single module, with a `choose_impl()` dispatcher that mirrors the `mss.linux` architecture. This prepares the project for future Windows backends (such as a DXGI-based implementation) that will sit alongside the existing GDI code.
+
+The existing GDI implementation has moved to `mss.windows.gdi`, but the previously-public names remain importable from `mss.windows` unchanged:
+
+- `mss.windows.MSS`
+- `mss.windows.MSSImplWindows`
+- `mss.windows.BACKENDS`
+
+Existing code does not need to be updated.
 
 ## macOS Stability Fix
 
--- a/src/mss/windows/__init__.py
+++ b/src/mss/windows/__init__.py
@@ -1,0 +1,63 @@
+"""Microsoft Windows backend dispatcher for MSS screenshot implementations."""
+
+from __future__ import annotations
+
+import warnings
+from typing import Any
+
+from mss.base import MSS as _MSS
+from mss.base import MSSImplementation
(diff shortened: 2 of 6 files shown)


BoboTiG:
Thanks @halldorfannar 🥂 

I'll take care of the documentation issues.

jholveck:
There's some other issues that make me uneasy here.  We'll fix them in a separate PR.

docs/source/release-history/v10.2.0.md
```diff
@@ -166,6 +170,18 @@ Additional improvements include:
 - improved Win32 error handling and diagnostics
 - fixes for capture failures during extended recordings
 
+### Package Restructure
+
+`mss.windows` is now a **package** rather than a single module, with a `choose_impl()` dispatcher that mirrors the `mss.linux` architecture. This prepares the project for future Windows backends (such as a DXGI-based implementation) that will sit alongside the existing GDI code.
```
jholveck, line 175: We don't need to mention the internal aspects: choose_impl, MSSImplWindows, BACKENDS, or mss.windows.gdi.  Discussing internal implementation details in user-facing documentation risks confusion about what users are supposed to access.

It should be sufficient to say that the Windows stuff was restructured, the existing GDI backend is still the default - and currently only - implementation, and that this was done to prepare for a DXGI implementation.

src/mss/windows/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+from __future__ import annotations
+
+import warnings
+from typing import Any
+
+from mss.base import MSS as _MSS
+from mss.base import MSSImplementation
+from mss.exception import ScreenShotError
+
+# Re-export the GDI implementation so existing code that references
```
jholveck, line 12: Unnecessary; MSSImplWindows was never publicly released.

src/mss/windows/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+# ``mss.windows.MSSImplWindows`` keeps working.
+from mss.windows.gdi import MSSImplWindows
+
+__all__ = ["MSS"]
+
+BACKENDS = ["default"]
```
jholveck, line 18: I would suggest adding an explicit "gdi", similar to how the Linux implementation has all the backends available by explicit request.  Currently, of course, this would be the same as "default".  The intention, though, is that "default" will eventually be able to look at things like the Windows version to choose the backend to use.

src/mss/windows/gdi.py
```diff
@@ -183,13 +163,14 @@ def _errcheck(result: BOOL | _Pointer, func: Callable, arguments: tuple) -> tupl
...
 class MSSImplWindows(MSSImplementation):
-    """Multiple ScreenShots implementation for Microsoft Windows.
+    """Multiple ScreenShots implementation for Microsoft Windows (GDI backend).
 
     This implementation uses CreateDIBSection for direct memory access to pixel data,
     which eliminates the need for GetDIBits. The DIB pixel data is written directly
     to system-managed memory that we can read from.
 
-    This has no Windows-specific constructor parameters.
+    This backend is selected by ``backend="default"`` and has no Windows-specific
```
jholveck, line 172: As above, I suggest making "gdi" the name for this backend, and "default" essentially an alias.

jholveck (commented):
There's just a couple of cleanups I'd suggest.

We don't need compatibility regarding MSSImplWindows.  Indeed, we certainly shouldn't advertise that, or other internal details, to users.

I'd also suggest naming the new backend "gdi" in the backend= parameter.  The goal for "default" is to have MSS select the backend, but each backend is still explicitly selectable.

You might consider updating the docs that mention backends, where they currently say that only Linux uses this facility, but I'm not sure if discussing Windows at this point is going to be insightful to end users.

PR #505, opened 2026-04-21
Follow up to windows backends

halldorfannar:

Address feedback from @jholveck and also some improvements to better align with backend strategy.

Also:
* fix test_sdist (`history.rst` replaces `changelog.rst`)
* fix .gitignore to properly exclude virtualenv
* fix pyproject.toml so that uv is happy (need to make docs build conditional on newer python due to sphinx 0.9)

- [x] Tests added/updated
- [x] Documentation updated
- [x] Changelog entry added
- [x] `./check.sh` passed


src/tests/bench_grab_windows.py
```diff
@@ -110,7 +110,6 @@ def benchmark_raw_bitblt() -> None:
     captureblt = 0x40000000
 
     with mss.MSS() as sct:
-        assert isinstance(sct._impl, mss.windows.MSSImplWindows)
```
jholveck, line 113: I think this may have caused the `quality` check to fail.

.gitignore
```diff
@@ -23,4 +23,4 @@ docs/output/
 .mypy_cache/
 __pycache__/
 ruff_cache/
-venv/
```
jholveck, line 26: I name my venv `venv`; can you leave that in `.gitignore` please?

docs/source/release-history/v10.2.0.md
```diff
@@ -170,17 +170,13 @@ Additional improvements include:
 - improved Win32 error handling and diagnostics
 - fixes for capture failures during extended recordings
 
-#### Package Restructure
+### Restructure
 
-`mss.windows` is now a **package** rather than a single module, with a `choose_impl()` dispatcher that mirrors the `mss.linux` architecture. This prepares the project for future Windows backends (such as a DXGI-based implementation) that will sit alongside the existing GDI code.
+The existing GDI implementation has moved to `mss.windows.gdi`. It is currently
```
jholveck: If you're going to mention the package name, go ahead and note that it's an internal package.  I really don't want to cause potential confusion on this issue.

.gitignore
halldorfannar: Absolutely!  You are the first person I meet who doesn't use the default 😅 I noticed that my .venv was getting picked up by git so that's why I made the change.

src/tests/bench_grab_windows.py
halldorfannar: Thanks for the pointer. I have been running `check` a lot but must have forgotten to do it after this last cleanup.

docs/source/release-history/v10.2.0.md
halldorfannar: Fixed.

docs/source/release-history/v10.2.0.md
```diff
@@ -170,17 +170,13 @@ Additional improvements include:
...
+to add a DXGI backend in the near future. 
 
-The existing GDI implementation has moved to `mss.windows.gdi`, but the previously-public names remain importable from `mss.windows` unchanged:
-
-- `mss.windows.MSS`
-- `mss.windows.MSSImplWindows`
-- `mss.windows.BACKENDS`
-
-Existing code does not need to be updated.
+These changes were made while keeping backwards compatibility with existing API.
```
jholveck:
```suggestion
These changes were made while keeping backwards compatibility with the existing API.
```

src/mss/windows/__init__.py
```diff
@@ -42,22 +38,26 @@ def choose_impl(backend: str = "default", **kwargs: Any) -> MSSImplementation:
...
-          ``CreateDIBSection`` for direct memory access to pixel data.
+        - ``"default"`` or ``"gdi"`` (default): GDI-based backend using
+          ``BitBlt`` and ``CreateDIBSection`` for direct memory access to pixel
+          data.
 
-        .. versionadded:: 10.3.0 Prior to this version, Windows had a single
-            implementation selected through :class:`mss.windows.MSS`.
+        .. versionadded:: 10.2.0 Prior to this version, this didn't exist:
+            Windows had a single implementation selected through
+            :class:`mss.windows.MSS`.
```
jholveck: This seems redundant with the versionadded block below.  Since the whole `choose_impl` function was added in 10.2.0, I'm not sure that it needs to be separately stated that the `backend` argument was also added at that time.

(I tried to get the suggestion to also delete the extra blank line, but couldn't get that to work.)

```suggestion
```

jholveck (approved):
Overall, this looks fine as-is.  I've added a couple tiny nits, but you can take them or leave them.

In `src/tests/conftest.py`, we do have a `backend` fixture that will iterate over the backends for the current platform.  This can be used explicitly by a test that needs to iterate over the backends.  It can also be used implicitly via the `mss_impl` fixture, which gives you an MSS factory, iterating over the backends.  Both of these mean that tests that use those will run multiple times, once for each backend.  When we do add the dxgi backend, you should fix the `backend` fixture accordingly.

docs/source/release-history/v10.2.0.md
halldorfannar: Done.

src/mss/windows/__init__.py
halldorfannar: Done.

BoboTiG:
Thanks @halldorfannar! :)

PR #507, opened 2026-04-22
Fix test_sdist for the changelog->history rename

jholveck:
- [X] Tests added/updated
- [X] `./check.sh` passed


jholveck:
Oops, I now see that the necessary fix was already committed.

PR #519, opened 2026-04-28
Improve error handling for Windows and MSS context

halldorfannar:

We essentially had three problems:

1. The error handling in Windows was slightly confusing where a Windows error object was created when it should not have been. This object was not being returned but it made log reading confusing.  We now have two separate error handling functions and only create the Windows error object when we truly have a last error from Windows.
2. The MSS context object was masking inner exceptions (from `with` scope) with outer exceptions (generated during `__exit__`) - this led to the wrong exception being surfaced by our tests. The MSS Context object has been enhanced to not mask the inner exception with an outer exception.  Once Python 3.11 becomes baseline we can use `ExceptionGroups` to surface both exceptions and still get the right outcome, but until then the common pattern I applied here will suffice.
3. The thread tests were not designed to propagate exceptions back to the main thread, which is required so that the test can report the correct failure.  This has been fixed and made available in a separate file so this mistake does not get repeated.

This relates to pull request #516. If you combine this with that you will see massive improvements in the error handling.

- [x] Tests added/updated
- [x] Documentation updated
- [x] Changelog entry added
- [x] `./check.sh` passed

### DIFF
--- a/src/mss/base.py
+++ b/src/mss/base.py
@@ -16,6 +16,7 @@ from mss.tools import to_png
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterator
+    from types import TracebackType
 
     from mss.models import Monitor, Monitors, Size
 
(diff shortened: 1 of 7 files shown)


src/mss/base.py
```diff
@@ -260,9 +261,21 @@ def __enter__(self) -> Self:
...
+        _exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        _traceback: TracebackType | None,
+    ) -> None:
         """For the cool call `with MSS() as mss:`."""
-        self.close()
+        try:
+            self.close()
+        except Exception:
+            # This extra work is needed so that exceptions generated during __exit__
```
jholveck, line 274: Wouldn't these normally be chained?  Is there a reason not to use chained exceptions here?  I'm a little hesitant to just swallow exceptions entirely.

halldorfannar: No. When a `with` body is already failing, and `__exit__` also fails during cleanup, chaining creates the wrong relationship and headline. Python/pytest reports the exception raised from `__exit__` as the active failure, so the real body failure becomes secondary context. That was the ReleaseDC masking the actual BitBlt/thread failure in the PR that triggered this work.

Python 3.11 has a new feature that is designed for this case, https://docs.python.org/3/library/exceptions.html#lib-exception-groups - but we cannot use that since we support 3.10.

BoboTiG:
Thanks @halldorfannar !

Issue #476, opened 2026-02-18
Proposed ScreenShot API changes


jholveck:
I’m proposing some API changes to ScreenShot that break compatibility in some ways, although common usages will still be fine.  This is to help us with our overall release planning, as being detailed in #471.

# Change 1: Buffers

## Goal

The goal here is to prevent unnecessary copies.  Currently, the API design requires us copy the pixel data in many conditions.

Many of these changes refer to the `Buffer` type.  This is in `collections.abc` as of Python 3.12, and also in `typing_extensions` if we want to use that for earlier versions.  We can alternatively alias it to `bytes | bytearray | memoryview` on earlier versions.

## Proposed API Changes

* `_raw`: Rename `raw` to `_raw`.  I’m unclear on whether this was ever intended to be exposed.  In the proposed new API, it’s the same to the user as `bgra` anyway, but making it private gives us more flexibility.
* `__init__`, `from_size`: Change the `data` argument from `bytearray` to `Buffer`.  This lets `_grab_impl` pass us a `memoryview` rather than having to make a new copy of the data.
* `bgra`, `rgb`: Return a read-only `Buffer` instead of a `bytes` object, avoiding a copy.

Also of note:
* `__array_interface__`: Not a type change, but I’m thinking we might make the provided array read-only.
* `buffer`, `__buffer__`: New method.  `__buffer__` is a Python 3.12-only feature that lets the ScreenShot object act as a Buffer itself, while the `buffer` method can be used on any version of Python.

## Impact

With just these simple API changes, and no backend-specific changes, the `cat-detector` and `video-capture` demos are 12% and 15% faster, respectively.  That’s for the whole program, not just the screen capture part.  The MSS code I used for this test is at <https://github.com/jholveck/python-mss/tree/d38af8a36abe2d7ce9f87f95bca78c7502f2a7ca>.  The proposed API further allows backends to avoid yet another copy, improving performance even more.

## Question: read-write or read-only?

One topic to discuss: currently, the user could modify the pixel data.  They had access to the read-write `bytearray` as `sct.raw`, and also could modify the raw data via NumPy using `__array_interface__`.

I’m thinking we may not want to make any guaranteed read-write mechanisms available, since that would mean we can’t cache `pixels` or `rgb`.  In my current implementation, the `.buffer()` method can be used to request a read-write buffer, since PyTorch ignores the read-only flag (with a warning).  But in general, I don’t see much sense in giving users a way to edit the buffer in-place.

[...]

PR #521, opened 2026-05-05
Change the ScreenShot API to be buffer-oriented.

jholveck:
This implements many of the changes discussed in #476, part 1.  It does not make the ScreenShot object itself accessible as a buffer; that's a separate matter.

This change makes `_raw` (previously `raw`) private.  `_raw` is now the original object passed in.

The `bgra` and `rgb` properties now return `memoryview` objects, rather than `bytes` or `bytearray` objects.  This means that the backends can now capture data to an OS-provided buffer, such as one provided by `mmap` or `CreateDIBSection`.  The `ScreenShot` object can then give the user direct access to this buffer, without ever needing to copy the data.

Similarly, ScreenShot objects now can be given their data as any buffer type, not just a bytearray.

The backends haven't yet been changed to take advantage of these new features.

Docs, examples, and demos are updated.

- [X] Tests added/updated
- [X] Documentation updated
- [X] Changelog entry added
- [X] `./check.sh` passed

### DIFF
--- a/demos/cat-detector.py
+++ b/demos/cat-detector.py
@@ -281,7 +281,7 @@ def main() -> None:
             # We transfer the image from MSS to PyTorch via a Pillow Image.  Faster approaches exist (see
             # screenshot_to_tensor), but PIL is more readable.  The bulk of the time in this program is spent doing
             # the AI work, so we just use the most convenient mechanism.
-            img = Image.frombytes("RGB", sct_img.size, sct_img.bgra, "raw", "BGRX")
+            img = Image.frombuffer("RGB", sct_img.size, sct_img.bgra, "raw", "BGRX")
 
             # We explicitly convert it to a tensor here, even though Torchvision can also convert it in the preprocess
             # step.  This is so that we send it to the GPU before we do the preprocessing: PIL Images are always on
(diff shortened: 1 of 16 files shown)


BoboTiG:
Thanks @jholveck :)

jholveck:
Sure thing!

As it stands, this already should improve performance: I didn't retest, but last I checked, it was about 10-20% faster overall in the demos.

Importantly, this lays the groundwork for the next change I've got in mind: zero-copy.  Once the OS gives us the screenshot data, we can give it to the users without ever having to copy it into a Python object.

That requires an extra stage, one that I've written to test but haven't yet put into MSS.  This is to couple the Python memory management, which knows when that data is no longer used, with the OS-supplied buffers' memory management.

As things stand after this commit, I think we can already change the way we use CreateDIBSection on Windows to improve things.  I've given Halldor my specific ideas on that.  But the real win is once I get the memory management stuff in, and the backends can start using it.

As I've said before, on my home machine, copying screenshot data for a 4k monitor takes about 3ms.  So minimizing those copies is a big performance win!

BoboTiG:
Yup, zero-copy will be such an improvement!

PR #533, opened 2026-06-01
Pin hatchling to 1.29.0

jholveck:
About an hour ago, a new version of our build system, Hatchling (1.30.0), was released.  This sets the metadata-version to 2.5.  But that version is not supported by Twine, our PyPI upload system.

I haven't researched if there's a better option, but this should deal with the immediate problem.  At the least, we should update Hatchling once Twine is able to handle 2.5.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling==1.29.0"]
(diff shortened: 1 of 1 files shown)


jholveck:
I forgot to mention in my commit notes: the new version of Hatchling caused our PyTest to start issuing failures in test_setup.py: it would ask Twine to ensure that the built artifacts are as expected.

PR #535, opened 2026-06-02
Fix Pillow examples using frombuffer

jholveck:
The Pillow docs recommend always using the decoder arguments when using frombuffer; see
https://pillow.readthedocs.io/en/stable/reference/Image.html

- [x] Tests added/updated - N/A
- [X] Documentation updated
- [X] Changelog entry added
- [X] `./check.sh` passed


jholveck:
Sorry, accidentally turned on the switch to automatically have Copilot check all my PRs.

PR #536, opened 2026-06-02
Make the data views writable

jholveck:
Previously (while working on 11.0), we changed the .bgra and .rgb attributes to be read-only memoryviews.  We also removed the .raw attribute.

The main reason we made the memoryviews read-only is so that we don't have to worry about cached values of .pixels and .rgb being in-sync with the underlying buffer, if users mutated the pixel data.

However, I now realize that this takes away a possible valuable use case: ctypes.  ctypes can only create an array (or pointer) from a buffer if it's writable; it doesn't support a read-only version. Previously, users could make a ctypes pointer using the .raw attribute.  The new API doesn't give them that ability, without copying the data (or using ctypes' from_address, which is perilous).

Read-only buffers also trigger a warning from PyTorch, although it's clearly-written and only is printed once.

This PR, as it stands, makes the `.bgra` and `.rgb` properties return writable memoryviews.  (It also restores `.raw`, this time as a deprecated alias of bgra, since there's no pressing need to remove the name entirely.)

It also documents that, while these memoryviews are writable, actually modifying the data may cause undefined behavior.

There's alternatives, of course.  Instead of what I've got here, ChatGPT instead suggests that we leave .bgra and .rgb read-only (as they were in 10.2, since they were `bytes` objects), and to add a property like `.writable_bgra` or something.  Its argument is as follows:

> The problem with making .bgra writable and saying “mutating is UB
> [undefined behavior]” is that Python users will absolutely see
> writable buffer and think “cool, in-place image editing.”  That’s
> not UB in the fun C sense; it’s more like “congratulations, you have
> a weird cache-invalidation footgun.”  The API shape would be lying a
> little.

I'm making this PR for discussion and consensus on the best way to go.

- [X] Tests added/updated
- [X] Documentation updated
- [x] Changelog entry added
- [X] `./check.sh` passed


jholveck:
@BoboTiG , @halldorfannar , do you have any comments on this, or should we just go ahead and merge it?

BoboTiG:
I think I prefer what you've done so far. I'm not really for the ChatGPT idea.

Documentation is clear enough, and I could live with UB :)

If @halldorfannar has no concern, OK to merge for me.

docs/source/release-history/v11.0.0.md
```diff
@@ -15,11 +15,13 @@ writing, but should be by the time we release 11.0!)
...
 
-The {py:attr}`mss.ScreenShot.raw` attribute has been removed.  Use the {py:attr}`mss.ScreenShot.bgra` property instead.
+The {py:attr}`mss.ScreenShot.raw` attribute has been deprecated, and will soon be removed.  Use the
+{py:attr}`mss.ScreenShot.bgra` property instead.
 
-The {py:attr}`mss.ScreenShot.bgra` and {py:attr}`mss.ScreenShot.rgb` properties now will return read-only bytes-like
+The {py:attr}`mss.ScreenShot.bgra` and {py:attr}`mss.ScreenShot.rgb` properties now will return bytes-like
 {py:type}`memoryview` objects, not necessarily {py:type}`bytes` or {py:type}`bytearray` objects.  For practical use
-cases, this should not be noticible.  This change was allows faster access to screenshot data, with fewer memory copies.
+cases, this should not be noticible.  This change was allows faster access to screenshot data, with fewer memory
```
halldorfannar:
```suggestion
cases, this should not be noticible.  This change allows faster access to screenshot data, with fewer memory
```

src/mss/screenshot.py
```diff
@@ -139,22 +162,33 @@ def rgb(self) -> memoryview:
         RGBRGB... sequence.  A specific pixel can be accessed as
         ``rgb[(y * width + x) * 4:(y * width + x) * 4 + 4].``
```
halldorfannar:
```suggestion
        ``rgb[(y * width + x) * 3:(y * width + x) * 3 + 3].``
```

halldorfannar (commented):
I'm also fine with this approach but found two documentation issues that I suggested fixes for.

jholveck:
Thanks for the feedback, both of you!

The current test failure is because of a minor issue in main, fixed by #549.  Once that gets committed to main, we can merge it into this branch, or we can just merge this into main directly (they don't conflict).

PR #566, opened 2026-08-03
Add immutable Monitor model

### LINKED ISSUE #470: Change Monitor dict to a proper class
I suggest we do this in the next major release.  Since Monitor now contains heterogenous types (int, bool, str) it feels quite messy to have it a dict.  When converting it to a class we could also make the attributes read-only, simply by having it inherit from a NamedTuple. I think this would be quite neat. But it's just a suggestion.

halldorfannar:

Fixes #470

Replace heterogeneous monitor dictionaries with a frozen, slotted `Monitor` dataclass that exposes typed geometry and
platform metadata attributes.

- Update the macOS, GNU/Linux, and Windows backends to produce and consume `Monitor` objects.
- Keep temporary string-key access for migration, while continuing to accept dictionaries and PIL-style tuples in
  `grab()`.
- Introduce new `Region` class and `as_region` factory function for Monitors; refactor code to use these
- Preserve immutable monitor descriptions in the video demos by constructing separate ojbects for cropped capture
  regions.
- Update filename formatting, examples, documentation, release notes, typing, and tests for the new model.
- Update AGENTS.md with instructions on how to update `test_sdist` when files are added, removed, or renamed.

This gives callers a stable, read-only monitor representation without preventing existing applications from passing
dictionary-based capture regions.

- [x] Tests added/updated
- [x] Documentation updated
- [x] Changelog entry added
- [x] `./check.sh` passed

Validation: `uv run ./check.sh`. The full pytest suite was not run during the final review.

### AI assistance disclosure

- [x] AI assistance was used to generate this contribution.

Codex assisted with reviewing the implementation, identifying and implementing follow-up fixes in the demos and
`ScreenShot` initialization, updating the immutability test for static typing, running validation, and drafting this PR
description. The contributor reviewed and directed the API and capture-region decisions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,11 +43,12 @@ Run tests:
 python -m pytest
 ```
 
-When adding, removing, or renaming a file included in the source distribution, update the explicit expected file list
-in `src/tests/test_setup.py::test_sdist` and run the focused packaging test:
+When adding, removing, or renaming a packaged file, update the explicit expected file list in
+`src/tests/test_setup.py::test_sdist` and, for files included in binary distributions, in
+`src/tests/test_setup.py::test_wheel`. Run the focused packaging tests:
 
 ```shell
-python -m pytest src/tests/test_setup.py::test_sdist
+python -m pytest src/tests/test_setup.py::test_sdist src/tests/test_setup.py::test_wheel
 ```
 
 On headless GNU/Linux environments, run tests with a virtual display:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,6 +43,13 @@ Run tests:
 python -m pytest
 ```
 
+When adding, removing, or renaming a file included in the source distribution, update the explicit expected file list
+in `src/tests/test_setup.py::test_sdist` and run the focused packaging test:
+
+```shell
+python -m pytest src/tests/test_setup.py::test_sdist
+```
+
 On headless GNU/Linux environments, run tests with a virtual display:
 
 ```shell
--- a/demos/cat-detector.py
+++ b/demos/cat-detector.py
@@ -224,7 +224,7 @@ def main() -> None:
         monitor = sct.monitors[1]
 
         # Compute the minimum size, in square pixels, that we'll consider reliable.
-        img_area = monitor["width"] * monitor["height"]
+        img_area = monitor.width * monitor.height
         min_box_area = MIN_AREA_FRAC * img_area
 
         # We start a new line of the log if the cat visibility status changes.  That way, your terminal will show
--- a/demos/tinytv-stream.py
+++ b/demos/tinytv-stream.py
@@ -148,6 +148,7 @@ from prettytable import PrettyTable, TableStyle
 from serial.tools import list_ports
 
 import mss
+from mss.models import Region
 
 from common.pipeline import Mailbox, PipelineStage
 
@@ -330,7 +331,7 @@ def _scale_stretch(img: Image.Image, size: tuple[int, int]) -> Image.Image:
 def capture_image(
     *,
     monitor: int | None = None,
-    capture_area: dict[str, int] | None = None,
+    capture_area: Region | None = None,
 ) -> Generator[Image.Image]:
     """Continuously capture images from the specified monitor.
 
@@ -338,13 +339,12 @@ def capture_image(
 
     :param monitor: Monitor number to capture from, using the standard
         MSS convention (all screens=0, first screen=1, etc.).
-    :param capture_area: Capture rectangle dict with 'left', 'top',
-        'width', 'height'.
+    :param capture_area: Capture region.
     :yields: PIL Image objects from the captured monitor.
     """
     with mss.MSS() as sct:
         rect = capture_area if capture_area is not None else sct.monitors[monitor]
-        LOGGER.debug("Capture area: %i,%i, %ix%i", rect["left"], rect["top"], rect["width"], rect["height"])
+        LOGGER.debug("Capture area: %i,%i, %ix%i", rect.left, rect.top, rect.width, rect.height)
 
         while True:
             sct_img = sct.grab(rect)
@@ -510,14 +510,14 @@ def _quality_type(value: str) -> int:
     raise argparse.ArgumentTypeError(msg)
 
 
-def _capture_area_type(value: str) -> dict[str, int]:
-    """Validate and return a capture area dict.
+def _capture_area_type(value: str) -> Region:
+    """Validate and return a capture region.
 
     Expected format is ``left,top,width,height`` where all values are
     integers.
 
     :param value: The capture area string to validate.
-    :returns: Dict with 'left', 'top', 'width', 'height' keys.
+    :returns: Capture region.
     :raises argparse.ArgumentTypeError: If the format is invalid or extents
         are non-positive.
     """
@@ -536,7 +536,7 @@ def _capture_area_type(value: str) -> dict[str, int]:
         msg = "Capture area width and height must be positive"
         raise argparse.ArgumentTypeError(msg)
 
-    return {"left": left, "top": top, "width": width, "height": height}
+    return Region(left=left, top=top, width=width, height=height)
 
 
 def main() -> None:
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -126,6 +126,7 @@ import av
 from si_prefix import si_format
 
 import mss
+from mss.models import Region
 
 from common.pipeline import Mailbox, PipelineStage
 
@@ -165,7 +166,7 @@ LOGGER = logging.getLogger("video-capture")
 def video_capture(
     fps: int,
     sct: mss.MSS,
-    capture_region: dict[str, int],
+    capture_region: Region,
     shutdown_requested: Event,
 ) -> Generator[tuple[mss.screenshot.ScreenShot, float], None, None]:
     # Keep track of the time when we want to get the next frame.  We limit the frame time this way instead of sleeping
@@ -436,12 +437,7 @@ def main() -> None:
     with mss.MSS() as sct:
         if args.region:
             left, top, right, bottom = args.region
-            capture_region = {
-                "left": left,
-                "top": top,
-                "width": right - left,
-                "height": bottom - top,
-            }
+            capture_region = Region(left=left, top=top, width=right - left, height=bottom - top)
         else:
             monitor = sct.monitors[args.monitor]
             capture_region = monitor.as_region()
@@ -454,8 +450,8 @@ def main() -> None:
             # it (at least, when using 4:2:0 subsampling).
             region_crop_to_multiple_of_two = codec in {"libx264", "libx265"}
         if region_crop_to_multiple_of_two:
-            capture_region["width"] = (capture_region["width"] // 2) * 2
-            capture_region["height"] = (capture_region["height"] // 2) * 2
+            capture_region.width = (capture_region.width // 2) * 2
+            capture_region.height = (capture_region.height // 2) * 2
 
         # We don't pass the container format to av.open here, so it will choose it based on the extension: .mp4, .mkv,
         # etc.
@@ -494,8 +490,8 @@ def main() -> None:
                 # so some video encoders will tag it as AVCOL_TRC_BT709 (1) instead.
                 video_stream.color_trc = 13
 
-            video_stream.width = capture_region["width"]
-            video_stream.height = capture_region["height"]
+            video_stream.width = capture_region.width
+            video_stream.height = capture_region.height
             # There are multiple time bases in play (stream, codec context, per-frame).  Depending on the container
             # and codec, some of these might be ignored or overridden.  We set the desired time base consistently
             # everywhere, so that the saved timestamps are correct regardless of what format we're saving to.
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -122,12 +122,10 @@ from typing import Any
 
 # Install the necessary libraries with "pip install av mss numpy si-prefix".
 import av
-import numpy as np
+from common.pipeline import Mailbox, PipelineStage
 from si_prefix import si_format
 
 import mss
-from common.pipeline import Mailbox, PipelineStage
-
 
 # These are the options you'd give to ffmpeg that it sends to the video codec.  Because ffmpeg and PyAV both use the
 # libav libraries, you can get the list of available flags with `ffmpeg -help encoder=libx264`, or whatever encoder
@@ -165,7 +163,7 @@ LOGGER = logging.getLogger("video-capture")
 def video_capture(
     fps: int,
     sct: mss.MSS,
-    monitor: mss.models.Monitor,
+    capture_region: dict[str, int],
     shutdown_requested: Event,
 ) -> Generator[tuple[mss.screenshot.ScreenShot, float], None, None]:
     # Keep track of the time when we want to get the next frame.  We limit the frame time this way instead of sleeping
@@ -183,7 +181,7 @@ def video_capture(
             time.sleep(next_frame_at - now)
 
         # Capture a frame, and send it to the next processing stage.
-        screenshot = sct.grab(monitor)
+        screenshot = sct.grab(capture_region)
         yield screenshot, now
 
         # We try to keep the capture rate at the desired fps on average.  If we can't quite keep up for a moment (such
@@ -436,7 +434,7 @@ def main() -> None:
     with mss.MSS() as sct:
         if args.region:
             left, top, right, bottom = args.region
-            monitor = {
+            capture_region = {
                 "left": left,
                 "top": top,
                 "width": right - left,
@@ -444,6 +442,12 @@ def main() -> None:
             }
         else:
             monitor = sct.monitors[args.monitor]
+            capture_region = {
+                "left": monitor.left,
+                "top": monitor.top,
+                "width": monitor.width,
+                "height": monitor.height,
+            }
 
         # Some codecs, such as libx264, require the region to be a multiple of 2, to get the chroma subsampling right.
         # Others, such as h264_nvenc, do not; they'll pad to get the subsampling region, and add flags to the stream
@@ -453,8 +457,8 @@ def main() -> None:
             # it (at least, when using 4:2:0 subsampling).
             region_crop_to_multiple_of_two = codec in {"libx264", "libx265"}
         if region_crop_to_multiple_of_two:
-            monitor["width"] = (monitor["width"] // 2) * 2
-            monitor["height"] = (monitor["height"] // 2) * 2
+            capture_region["width"] = (capture_region["width"] // 2) * 2
+            capture_region["height"] = (capture_region["height"] // 2) * 2
 
         # We don't pass the container format to av.open here, so it will choose it based on the extension: .mp4, .mkv,
         # etc.
@@ -493,8 +497,8 @@ def main() -> None:
                 # so some video encoders will tag it as AVCOL_TRC_BT709 (1) instead.
                 video_stream.color_trc = 13
 
-            video_stream.width = monitor["width"]
-            video_stream.height = monitor["height"]
+            video_stream.width = capture_region["width"]
+            video_stream.height = capture_region["height"]
             # There are multiple time bases in play (stream, codec context, per-frame).  Depending on the container
             # and codec, some of these might be ignored or overridden.  We set the desired time base consistently
             # everywhere, so that the saved timestamps are correct regardless of what format we're saving to.
@@ -535,7 +539,7 @@ def main() -> None:
                     video_capture,
                     fps,
                     sct,
-                    monitor,
+                    capture_region,
                     shutdown_requested,
                 ),
                 out_mailbox=mailbox_screenshot,
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -121,6 +121,7 @@ from threading import Event
 from typing import Any
 
 # Install the necessary libraries with "pip install av mss numpy si-prefix".
+# Note that numpy is needed even if we don't explicitly import it here.
 import av
 from si_prefix import si_format
 
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -122,10 +122,11 @@ from typing import Any
 
 # Install the necessary libraries with "pip install av mss numpy si-prefix".
 import av
-from common.pipeline import Mailbox, PipelineStage
 from si_prefix import si_format
 
 import mss
+
+from common.pipeline import Mailbox, PipelineStage
 
 # These are the options you'd give to ffmpeg that it sends to the video codec.  Because ffmpeg and PyAV both use the
 # libav libraries, you can get the list of available flags with `ffmpeg -help encoder=libx264`, or whatever encoder
@@ -442,12 +443,7 @@ def main() -> None:
             }
         else:
             monitor = sct.monitors[args.monitor]
-            capture_region = {
-                "left": monitor.left,
-                "top": monitor.top,
-                "width": monitor.width,
-                "height": monitor.height,
-            }
+            capture_region = monitor.as_capture_region()
 
         # Some codecs, such as libx264, require the region to be a multiple of 2, to get the chroma subsampling right.
         # Others, such as h264_nvenc, do not; they'll pad to get the subsampling region, and add flags to the stream
--- a/demos/video-capture.py
+++ b/demos/video-capture.py
@@ -444,7 +444,7 @@ def main() -> None:
             }
         else:
             monitor = sct.monitors[args.monitor]
-            capture_region = monitor.as_capture_region()
+            capture_region = monitor.as_region()
 
         # Some codecs, such as libx264, require the region to be a multiple of 2, to get the chroma subsampling right.
         # Others, such as h264_nvenc, do not; they'll pad to get the subsampling region, and add flags to the stream
--- a/docs/source/examples/custom_cls_image.py
+++ b/docs/source/examples/custom_cls_image.py
@@ -7,7 +7,7 @@ Screenshot of the monitor 1, using a custom class to handle the data.
 from typing import Any
 
 import mss
-from mss.models import Monitor
+from mss.models import CaptureRegion
 from mss.screenshot import ScreenShot
 
 
@@ -17,9 +17,9 @@ class SimpleScreenShot(ScreenShot):
     or add new methods.
     """
 
-    def __init__(self, data: bytearray, monitor: Monitor, **_: Any) -> None:
+    def __init__(self, data: bytearray, region: CaptureRegion, **_: Any) -> None:
         self.data = data
-        self.monitor = monitor
+        self.region = region
 
 
 with mss.MSS() as sct:
--- a/docs/source/examples/custom_cls_image.py
+++ b/docs/source/examples/custom_cls_image.py
@@ -7,7 +7,7 @@ Screenshot of the monitor 1, using a custom class to handle the data.
 from typing import Any
 
 import mss
-from mss.models import CaptureRegion
+from mss.models import Region
 from mss.screenshot import ScreenShot
 
 
@@ -17,7 +17,7 @@ class SimpleScreenShot(ScreenShot):
     or add new methods.
     """
 
-    def __init__(self, data: bytearray, region: CaptureRegion, **_: Any) -> None:
+    def __init__(self, data: bytearray, region: Region, **_: Any) -> None:
         self.data = data
         self.region = region
 
--- a/docs/source/release-history/v10.2.0.md
+++ b/docs/source/release-history/v10.2.0.md
@@ -260,7 +260,7 @@ In 11.0, monitor dictionaries will become a dedicated **`Monitor` class**.
 
 To maintain compatibility:
 
-- dictionary-style access will continue to work
+- string-key access will temporarily continue to work
 
 ```python
 monitor["left"]
@@ -268,6 +268,14 @@ monitor["top"]
 ```
 
 - `grab()` will continue accepting dictionaries
+
+The compatibility access does not make `Monitor` a complete mapping.  Migrate dictionary methods, membership tests, and
+unpacking to attribute access:
+
+```python
+monitor.left
+monitor.top
+```
 
 If you use type annotations, you can switch to the provided `Monitor` type:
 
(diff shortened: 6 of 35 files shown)


demos/video-capture.py
```diff
@@ -122,12 +122,10 @@
 
 # Install the necessary libraries with "pip install av mss numpy si-prefix".
 import av
-import numpy as np
+from common.pipeline import Mailbox, PipelineStage
```
jholveck: I'd rather not put that here, with the third-party stuff, don't you think?

demos/video-capture.py
```diff
@@ -436,14 +434,20 @@ def main() -> None:
...
-            monitor = {
+            capture_region = {
                 "left": left,
                 "top": top,
                 "width": right - left,
                 "height": bottom - top,
             }
         else:
             monitor = sct.monitors[args.monitor]
+            capture_region = {
```
jholveck: This makes me think that we may want to make something that converts the monitor to a capture region, like `my_monitor.as_capture_region()` or the like.

docs/source/release-history/v11.0.0.md
```diff
@@ -22,6 +22,24 @@ The {py:attr}`mss.ScreenShot.bgra` and {py:attr}`mss.ScreenShot.rgb` properties
 {py:class}`memoryview` objects, rather than {py:class}`bytes` or {py:class}`bytearray` objects.  For practical use
 cases, this should not be noticeable.  This change allows faster access to screenshot data, with fewer memory copies.
 
+#### Immutable monitor objects
+
+{py:attr}`mss.MSS.monitors` now returns frozen, slotted {py:class}`mss.models.Monitor` objects instead of dictionaries.
```
jholveck: Why frozen?  (I was going to ask about that in your latest proposal.)

src/mss/linux/base.py
```diff
@@ -305,24 +306,31 @@ def _monitors_from_randr_monitors(
...
+
+            monitors.append(
+                Monitor(
+                    left=randr_monitor.x,
+                    top=randr_monitor.y,
+                    width=randr_monitor.width,
+                    height=randr_monitor.height,
+                    # Under XRandR, it's legal for no monitor to be primary.  In
+                    # this case, case MSSBase.primary_monitor will return the
+                    # first monitor.  That said, we note in the dict that we
```
jholveck:
```suggestion
                    # first monitor.  That said, we note in the Monitor that we
```

src/mss/linux/base.py
```diff
@@ -305,24 +306,31 @@ def _monitors_from_randr_monitors(
...
+                    # explicitly are told by XRandR that all of the monitors are
+                    # not primary.  (This is distinct from the XRandR 1.2 path,
+                    # which doesn't have any information about primary
+                    # monitors.)
+                    is_primary=bool(randr_monitor.primary),
+                    name=output_ids.get("name"),
+                    unique_id=output_ids.get("unique_id"),
+                    output=output_ids.get("output"),
```
jholveck: Maybe `**output_ids` instead, to avoid future mistakes if one place is updated but not all three (`_randr_output_ids`, `_monitors_from_randr_monitors`, and `_monitors_from_randr_crtcs`)?

src/mss/linux/base.py
```diff
@@ -352,23 +360,24 @@ def _monitors_from_randr_crtcs(
...
+            monitors.append(
+                Monitor(
+                    left=crtc_info.x,
+                    top=crtc_info.y,
+                    width=crtc_info.width,
+                    height=crtc_info.height,
+                    is_primary=chosen_output == primary_output if primary_output is not None else None,
+                    name=output_ids.get("name"),
+                    unique_id=output_ids.get("unique_id"),
+                    output=output_ids.get("output"),
```
jholveck: As above

src/mss/base.py
```diff
@@ -303,14 +304,25 @@ def grab(self, monitor: Monitor | tuple[int, int, int, int], /) -> ScreenShot:
...
-
-        if monitor["width"] <= 0 or monitor["height"] <= 0:
+            monitor = Monitor(
+                left=monitor[0],
+                top=monitor[1],
+                width=monitor[2] - monitor[0],
+                height=monitor[3] - monitor[1],
+            )
+        elif isinstance(monitor, dict):
+            monitor = Monitor(
```
jholveck: Why are we converting dict capture regions into `Monitor` objects?  I'm having a hard time seeing why we don't normalize everything to regions and pass those down to `_impl.grab`, rather than making these synthetic monitor objects.

This is the biggest concern I have with this PR.  (I didn't call this out in all of the `grab` implementations, to avoid duplication.)

src/mss/screenshot.py
```diff
@@ -35,15 +35,25 @@ class ScreenShot:
...
     __slots__ = {"__bgra", "__pixels", "__rgb", "_raw", "pos", "size"}
 
-    def __init__(self, data: Buffer, monitor: Monitor, /, *, size: Size | None = None) -> None:
+    def __init__(self, data: Buffer, monitor: Monitor | dict[str, Any], /, *, size: Size | None = None) -> None:
         self.__pixels: Pixels | None = None
         self.__rgb: memoryview | None = None
 
+        if isinstance(monitor, Monitor):
+            left, top = monitor.left, monitor.top
+        else:  # remove this once we deprecate the string key access
```
jholveck: Deprecate it where?  Why does this allow a dict to be passed to the constructor?

This relates to https://github.com/BoboTiG/python-mss/pull/566/changes#r3709864024

AGENTS.md
```diff
@@ -43,6 +43,13 @@ Run tests:
 python -m pytest
 ```
 
+When adding, removing, or renaming a file included in the source distribution, update the explicit expected file list
```
jholveck: If you're going to mention this, you may as well also mention test_wheel for files that make it into the binary distributions.

jholveck (commented):
I'm not really comfortable with this idea of making synthetic Monitor objects to send down through `_impl.grab`, rather than sending capture regions as dicts or tuples.  Can you elaborate on why you made that choice?

demos/video-capture.py
halldorfannar: Done, moved it back.

demos/video-capture.py
halldorfannar: Yes, wasn't sure which way you guys wanted to go here. I have added that function and changed the code.

docs/source/release-history/v11.0.0.md
halldorfannar: I have added documentation to explain why.  It is important that customers cannot modify the monitors because they are owned by MSS and should be immutable.

src/mss/linux/base.py
halldorfannar: Fixed.

src/mss/linux/base.py
halldorfannar: Good idea. Implemented.

src/mss/base.py
halldorfannar: Fixed with the new `as_capture_region`.

src/mss/screenshot.py
halldorfannar: My understanding was that during transition we wanted to support the key string access but eventually we will remove it.  Just to soften the blow for existing customers and their code.  We will mark this for deprecation and eventually remove it.  My language is slightly off in the comment.  We are marking it for deprecation and will later remove it.  The process is normally:

1. Deprecate
2. At a later date remove

*BUT* now that we have moved to the CaptureRegion this code has all been thrown away and we will remain backwards compatible (no deprecation planned anymore).

AGENTS.md
halldorfannar: Good point.  I have added that.

docs/source/release-history/v11.0.0.md
jholveck: To clarify: I hid interpreted "frozen" in the sense of C++'s final, or possibly making some sort of future compatibility guarantee, but now that I see your latest commit, I see you meant it in the sense of "immutable".

Might that be a better term?

(I'm also not sure on why it matters to users that they're slotted, but that's not a problem.)

halldorfannar:
One overall thing. I'm wondering if the name `CaptureRegion` is too limiting.  Maybe `Region` is better and more futureproof.  It doesn't matter what the region will be used for inside MSS.  It will always have the same coordinates associated with it.  Using it directly for capture is just one application I can think of.

src/mss/screenshot.py
jholveck: This ties into my earlier question about why we pass around these synthetic `Monitor` objects wherever we're talking about capture regions.  Unfortunately, the link I got for that didn't work as expected.

At this point, when I wrote the review, the things that construct a `ScreenShot` - the `_impl.grab` methods - would always pass it a `Monitor` object, never a dict.

Also, this code would be in the context of allowing the caller to pass in a `Monitor` or a dict, but the comment about "string key access" ties into how a user can access the properties of a `Monitor`, which is a completely different matter.

But now that we're using `CaptureRegion`, then it's moot.

jholveck:
> One overall thing. I'm wondering if the name `CaptureRegion` is too limiting. Maybe `Region` is better and more futureproof. It doesn't matter what the region will be used for inside MSS. It will always have the same coordinates associated with it. Using it directly for capture is just one application I can think of.

Yeah, I think that `Region` may be better.  For instance, we might use it in a `Window` object to give the window geometry to the user.

(Besides, `as_capture_region` felt a bit verbose; it was just the first thing that occurred to me when I wrote that comment.)

Food for thought: should a `Region` carry with it the metric space (nominal vs. physical)?  Probably not practically, and definitely not in this PR, but it may be at least worth passing through the brain once.

docs/source/release-history/v11.0.0.md
halldorfannar: `frozen` and `slots` are keywords from Python that I feel we should use here, because there are still ways for the user to sneakily avoid the immutability achieved by these.  In the next line we say:

```
Monitor entries are cached display-configuration snapshots; keeping them immutable prevents application code from
accidentally changing the geometry or metadata held by an {py:class}`mss.MSS` instance.
```

Which I feel makes it quite clear.  I could also say `making them harder to mutate` instead of `keeping them immutable`. But maybe I'm not seeing something that you are seeing.

src/mss/screenshot.py
halldorfannar: Agreed. Resolving.

docs/source/release-history/v11.0.0.md
halldorfannar: I have changed the text in alignment with what we discussed during our 1:1.

halldorfannar:
This is ready to merge.

src/mss/screenshot.py
```diff
@@ -35,15 +37,18 @@ class ScreenShot:
...
     __slots__ = {"__bgra", "__pixels", "__rgb", "_raw", "pos", "size"}
 
-    def __init__(self, data: Buffer, monitor: Monitor, /, *, size: Size | None = None) -> None:
+    def __init__(self, data: Buffer, region: Region, /, *, size: Size | None = None) -> None:
         self.__pixels: Pixels | None = None
         self.__rgb: memoryview | None = None
 
         #: NamedTuple of the screenshot coordinates.
-        self.pos: Pos = Pos(monitor["left"], monitor["top"])
+        self.pos: Pos = Pos(region["left"], region["top"])
```
BoboTiG:
```suggestion
        self.pos: Pos = Pos(region.left, region.top)
```

Should we use the preferred syntax internally everywhere?

halldorfannar: This is a good point. The current class as I've designed it, doesn't support that - but I think it should.  I made it a TypedDict to make my changes easier to implement but I think what you propose is better. Let's make it a proper attribute class and I will just keep dict support as a special backwards compat in the functions that need it. 

I will make the new attribute style access the preferred way in our internal code.

halldorfannar: OK, the relevant changes have been made.

BoboTiG (approved):

demos/cat-detector.py
```diff
@@ -224,7 +224,7 @@ def main() -> None:
         monitor = sct.monitors[1]
```
jholveck, line 224: Now that we have primary_monitor, we should use that.
```suggestion
        monitor = sct.primary_monitor
```

demos/tinytv-stream.py
```diff
@@ -510,14 +510,14 @@ def _quality_type(value: str) -> int:
...
-    """Validate and return a capture area dict.
+def _capture_area_type(value: str) -> Region:
+    """Validate and return a capture region.
 
     Expected format is ``left,top,width,height`` where all values are
     integers.
 
     :param value: The capture area string to validate.
-    :returns: Dict with 'left', 'top', 'width', 'height' keys.
+    :returns: Capture region.
```
jholveck, line 520: I normally don't leave in `:returns:` if it's obvious from the context and return type.  I do for dicts, to clarify the keys, but I don't think it's needed for `Region`.

```suggestion
```

docs/source/examples/custom_cls_image.py
jholveck, line 27: While we're in here...

```suggestion
    image = sct.grab(sct.primary_monitor)
```

docs/source/release-history/v10.2.0.md
```diff
@@ -260,7 +260,7 @@ In 11.0, monitor dictionaries will become a dedicated **`Monitor` class**.
 
 To maintain compatibility:
 
-- dictionary-style access will continue to work
+- string-key access will temporarily continue to work
```
jholveck, line 263: Just to make it clear even to new programmers:

```suggestion
- string-key (dictionary-style) access will temporarily continue to work
```

docs/source/release-history/v11.0.0.md
```diff
@@ -22,6 +22,31 @@ The {py:attr}`mss.ScreenShot.bgra` and {py:attr}`mss.ScreenShot.rgb` properties
...
+#### Immutable monitor objects
+
+{py:attr}`mss.MSS.monitors` now returns immutable {py:class}`mss.models.Monitor`
+objects instead of dictionaries. Monitor entries are cached
+display-configuration snapshots; making them frozen and slotted prevents
+application code from accidentally changing the geometry or metadata held by an
+{py:class}`mss.MSS` instance. Use attributes to read geometry and metadata:
+
+```python
+monitor = sct.monitors[1]
```
jholveck, line 34: Just to get people in the habit of thinking about `primary_monitor` instead of `monitors[1]`, I suggest we use that.

```suggestion
monitor = sct.primary_monitor  # or monitors[1], etc.
```

src/mss/linux/base.py
```diff
@@ -25,6 +26,12 @@
 ALL_PLANES = 0xFFFFFFFF  # XCB doesn't define AllPlanes
 
 
+class _RandROutputIds(TypedDict, total=False):
```
jholveck, line 29: I might name this `_RandROutputInfo` or something instead, since it's not the ID.  In fact, it doesn't even include the XID that XRandR uses.  (The `unique_id` is something different, a URL encoding of certain EDID data.)

src/mss/linux/base.py
```diff
@@ -305,24 +312,29 @@ def _monitors_from_randr_monitors(
...
+                output_ids = self._randr_output_ids(chosen_output, timestamp, edid_atom)
+
+            monitors.append(
+                Monitor(
+                    left=randr_monitor.x,
+                    top=randr_monitor.y,
+                    width=randr_monitor.width,
+                    height=randr_monitor.height,
+                    # Under XRandR, it's legal for no monitor to be primary.  In
+                    # this case, case MSSBase.primary_monitor will return the
```
jholveck, line 328: We can fix this typo that had been in the original:

```suggestion
                    # this case, MSSBase.primary_monitor will return the
```

src/mss/linux/base.py
```diff
@@ -424,13 +435,13 @@ def cursor(self) -> ScreenShot:
...
-        """Retrieve pixels from a monitor using ``GetImage``.
+    def _grab_xgetimage(self, region: Region, /) -> bytearray:
+        """Retrieve pixels from a capture region using ``GetImage``.
 
         Used by the XGetImage backend and by the XShmGetImage backend in
         fallback mode.
 
-        :param monitor: Monitor rectangle specifying ``left``, ``top``,
+        :param region: Rectangle specifying ``left``, ``top``,
             ``width``, and ``height`` to capture.
```
jholveck, line 445: This phrasing still has a bit of a dict-flavored tone.

```suggestion
        :param region: Desktop area to capture.
```

src/mss/base.py
```diff
@@ -291,37 +296,47 @@ def close(self) -> None:
...
 
-        Note: ``monitor`` can be a tuple like the one
-        :py:func:`PIL.ImageGrab.grab` accepts: ``(left, top, right, bottom)``
+        ``region`` can be a :class:`mss.models.Monitor`, a :class:`mss.models.Region`, a region dictionary, or a tuple
+        like the one :py:func:`PIL.ImageGrab.grab` accepts: ``(left, top, right, bottom)``.
 
-        :param monitor: The coordinates and size of the box to capture.
-                        See :meth:`monitors <monitors>` for object details.
+        :param region: The coordinates and size of the box to capture.
+                       See :meth:`monitors <monitors>` for monitor object details.
```
jholveck, line 306: This reference to `monitors` can probably be removed.  It was needed when the required dict format wasn't specified elsewhere, but now the `Monitor` class has a docstring explaining its fields.

```suggestion
```

src/mss/base.py
```diff
@@ -291,37 +296,47 @@ def close(self) -> None:
...
-
-        if monitor["width"] <= 0 or monitor["height"] <= 0:
-            msg = f"Region has zero or negative size: {monitor!r}"
+        if isinstance(region, tuple):
+            grab_region = Region(
+                left=region[0],
+                top=region[1],
+                width=region[2] - region[0],
+                height=region[3] - region[1],
+            )
```
jholveck, line 315: We may want to do validation and element type coercion here.

```suggestion
        if isinstance(region, tuple):
            if len(region) != 4:  # noqa: PLR2004
                msg = "Tuples for grab must have exactly four elements: left, top, right, bottom"
                raise ScreenShotError(msg)
            grab_region = Region(
                left=int(region[0]),
                top=int(region[1]),
                width=int(region[2] - region[0]),
                height=int(region[3] - region[1]),
            )
```

src/mss/base.py
```diff
@@ -291,37 +296,47 @@ def close(self) -> None:
...
+        elif isinstance(region, Monitor):
+            grab_region = region.as_region()
+        elif isinstance(region, Region):
+            grab_region = copy(region)
```
jholveck, line 319:
```suggestion
        elif isinstance(region, Region):
            # Make a copy, in case the user changes the Region object later.  Also, coerce the elements to ints.
            grab_region = Region(
                left=int(region.left),
                top=int(region.top),
                width=int(region.width),
                height=int(region.height),
            )
```

src/mss/base.py
```diff
@@ -291,37 +296,47 @@ def close(self) -> None:
...
+        elif isinstance(region, dict):
+            grab_region = Region(
+                left=region["left"],
+                top=region["top"],
+                width=region["width"],
+                height=region["height"],
+            )
+
```
jholveck, line 327: I would recommend adding a default clause.  Do we want this to be a `ScreenShotError` or a `TypeError`?

```suggestion
        else:
            msg = f"Capture area must be a Region, Monitor, tuple, or dict: {grab_region!r}"
            raise ScreenShotError(msg)

```

src/mss/base.py
```diff
@@ -335,19 +350,19 @@ def monitors(self) -> Monitors:
...
         - ``height``: the height
-        - ``is_primary``: (optional) true if this is the primary monitor
-        - ``name``: (optional) human-readable device name
-        - ``unique_id``: (optional) platform-specific stable identifier for the monitor
-        - ``output``: (optional, Linux only) monitor output name, compatible with xrandr
+        - ``is_primary``: true or false when known, otherwise ``None``
+        - ``name``: human-readable device name, or ``None``
+        - ``unique_id``: platform-specific stable identifier, or ``None``
+        - ``output``: Linux output name compatible with xrandr, or ``None``
         """
```
jholveck, line 366: I would suggest making this docstring more user-facing.  The existing wording is geared towards implementers.  It's also mostly a relic from from before we changed to a strategy design, with its expectation that the implementation-specific subclass would populate `self._monitors`.

```suggestion
    @property
    def monitors(self) -> Monitors:
        """Get positions of all monitors.

        If a monitor is rotated, its dimensions reflect its displayed
        orientation. In other words, height is the visible height, not
        the number of scanout lines.

        The first element, ``monitors[0]``, is all the monitors
        together.  It is the virtual desktop, holding all the monitors.

        The remaining elements, ``monitors[N]`` (with N > 0), are the
        individual displays.

        If monitors are being mirrored, the list will only include one
        of the mirrored copies.

        .. seealso::
            :py:attr:`primary_monitor`
        """
```

src/mss/base.py
```diff
@@ -425,7 +453,18 @@ def save(
...
+                date=datetime.now(UTC) if "{date" in output else None,
+                top=monitor.top,
+                left=monitor.left,
+                width=monitor.width,
+                height=monitor.height,
+                is_primary=monitor.is_primary,
+                name=monitor.name,
+                unique_id=monitor.unique_id,
+                output=monitor.output,
+            )
```
jholveck, line 467: I might suggest that we use one constant date for all of the monitors.

I'd also not try to check for `{date`: not checking the date may be marginally faster if not needed, but we're also dealing with false positives if somebody uses `{{date` or some such.

Finally, I might suggest that we use use the local time instead of UTC, since that seems more likely to be useful to the end user, although admittedly I guess nobody's complained.  In this suggestion, I also use `astimezone` to add the tzinfo to the datetime object so that `{date:%Y-%m-%dT%H:%M:%S.%f%:z}` still be able to use tzinfo for the `%:z` component.

```suggestion

        date = datetime.now().astimezone()
        if mon == 0:
            # One screenshot by monitor
            for idx, monitor in enumerate(monitors[1:], 1):
                fname = output.format(
                    mon=idx,
                    date=date,
                    top=monitor.top,
                    left=monitor.left,
                    width=monitor.width,
                    height=monitor.height,
                    is_primary=monitor.is_primary,
                    name=monitor.name,
                    unique_id=monitor.unique_id,
                    output=monitor.output,
                )
                if callable(callback):
                    callback(fname)
                sct = self.grab(monitor)
                to_png(sct.rgb, sct.size, level=self.compression_level, output=fname)
                yield fname
        else:
            # A screenshot of all monitors together or
            # a screenshot of the monitor N.
            mon = 0 if mon == -1 else mon
            try:
                monitor = monitors[mon]
            except IndexError as exc:
                msg = f"Monitor {mon!r} does not exist."
                raise ScreenShotError(msg) from exc

            output = output.format(
                mon=mon,
                date=date,
                top=monitor.top,
                left=monitor.left,
                width=monitor.width,
                height=monitor.height,
                is_primary=monitor.is_primary,
                name=monitor.name,
                unique_id=monitor.unique_id,
                output=monitor.output,
            )
```

src/mss/models.py
```diff
@@ -2,11 +2,68 @@
...
+    """
+
+    left: int
+    top: int
+    width: int
+    height: int
+    is_primary: bool | None = None
+    name: str | None = None
+    unique_id: str | None = None
+    output: str | None = None
```
jholveck, line 42: I suggest we put the attribute docstrings on the relevant attributes, using Sphinx's `#:` syntax.  Here, I'm also editing the docstrings to provide a bit more information, and clarify how persistent `unique_id` is.

```suggestion
    """Monitor geometry and optional platform metadata.

    .. seealso::
       - :py:attr:`.MSS.monitors`
       - :py:attr:`.MSS.primary_monitor`

    .. version-changed:: 11.0.0
       Prior to this version, ``Monitor`` was an alias for ``dict[str, int]``.
       In MSS 11, it is still possible to access attributes with dict-style
       string-key access, such as ``monitor["left"]``, but this
       behavior is deprecated and will be removed in a later version.
    """

    #: The monitor's left edge within the entire virtual desktop.
    left: int
    #: The monitor's top edge within the entire virtual desktop.
    top: int
    width: int
    height: int
    #: The human-readable name of this monitor, typically the brand
    #: and model.
    #: .. version-added:: 10.2.0
    name: str | None = None
    #: Whether this is the primary monitor, according to the operating
    #: system.  If MSS can't determine the primary monitor, this will
    #: be ``None`` for all monitors, although
    #: :py:attr:`.MSS.primary_monitor` will still return a monitor
    #: (the first one).
    #: .. version-added:: 10.2.0
    is_primary: bool | None = None
    #: The platform-specific stable identifier.  This is
    #: generally stable across reboots, or ordinary disconnection /
    #: reconnection, but may change when the display hardware or
    #: connection topology changes.
    #: .. version-added:: 10.2.0
    unique_id: str | None = None
    #: The short output name, for interfacing with other tools.  This
    #: is only currently populated by Linux, where it is the name used
    #: by xrandr.
    #: .. version-added:: 10.2.0
    output: str | None = None
```

src/tests/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+
+from collections.abc import Mapping
+from dataclasses import FrozenInstanceError
+
+import pytest
+
+from mss.models import Monitor, Region
+
+
+def test_region() -> None:
```
jholveck, line 11: I question the utility of several of the tests in this file, but I'm not going to say that they need to be deleted.

src/tests/test_save.py
```diff
@@ -60,10 +60,36 @@ def test_output_format_positions_and_sizes(mss_impl: Callable[..., MSS]) -> None
...
+    def capture_filename(value: str) -> None:
+        nonlocal filename
+        filename = value
+        raise FormattingCompleteError
+
+    fmt = "sct-{is_primary}-{unique_id}.png"
+    with mss_impl() as sct:
+        monitor = sct.monitors[1]
+        with pytest.raises(FormattingCompleteError):
+            next(sct.save(mon=1, output=fmt, callback=capture_filename))
```
jholveck, line 88: Seeing this `next` makes me wonder: is there a reason that we make `save` a generator, instead of saving all the screenshots immediately and returning a list?  I looked at the commit history, and this actually goes back to the very first MSS code, in 237a744fa80eed0a7e0df6955cc663a031a32ae6.  But it does seem a bit surprising to me that we're doing it this way.

If we want to change it, I can open a new issue.

docs/source/release-history/v10.2.0.md
```diff
@@ -269,6 +269,14 @@ monitor["top"]
...
 - `grab()` will continue accepting dictionaries
 
+The compatibility access does not make `Monitor` a complete mapping.  Migrate dictionary methods, membership tests, and
+unpacking to attribute access:
+
+```python
+monitor.left
+monitor.top
+```
+
```
jholveck, line 279: I might suggest that we either list all the attributes, or even just the four coordinate ones, or if we want a brief list, just put them inline.  I think that just a couple inline would be fine here.

```suggestion
The compatibility access does not make `Monitor` a complete mapping.  Migrate dictionary methods, membership tests, and
unpacking to attribute access, such as ``monitor.left`` and ``monitor.top``.
```

docs/source/usage.rst
```diff
@@ -52,10 +52,30 @@ list of all the monitors, starting from index 1, as well as the full virtual scr
...
 
-For capturing a specific region, you can pass :py:meth:`mss.MSS.grab` a dictionary with the keys ``top``, ``left``,
-``width``, and ``height``.  For instance, to capture a 100x100 pixel region starting at the top-left corner of the
-screen, you could use :python:`{"top": 0, "left": 0, "width": 100, "height": 100}`.  You can also use a PIL-style box,
-which is a 4-tuple of ``(left, top, right, bottom)``.
+Each entry is an immutable :py:class:`mss.models.Monitor`.  Its geometry is available through ``left``, ``top``,
+``width``, and ``height`` attributes.  The ``is_primary``, ``name``, ``unique_id``, and ``output`` metadata attributes
+are ``None`` when unavailable::
+
+    monitor = sct.monitors[1]
```
jholveck, line 59:
```suggestion
    monitor = sct.primary_monitor
```

docs/source/usage.rst
```diff
@@ -52,10 +52,30 @@ list of all the monitors, starting from index 1, as well as the full virtual scr
...
+    print(monitor.width, monitor.height)
+
+Call ``monitor.as_region()`` when you need its geometry as a :py:class:`mss.models.Region`::
```
jholveck, line 62: This "Capturing Screenshots" section is meant to be a "how to get started" introduction to MSS.  It tries to focus on the essentials, and avoid confusing new users.

With that in mind, most of this seems to be just an unnecessary distraction.  It increases the cognitive load on new users.  For instance, are new users who are just learning the basics likely to care that `**monitor` unpacking is not supported?  It's just extra stuff that they don't have to care about, but still have to spend cognitive cycles to decide if they care.

For the purposes of this flow, I think that explaining all about a `Monitor` object may be a bit too much.  I might suggest putting the details in a separate section or subsection, or even just using some cross-references into the API reference for some details.  Putting some of this into a separate section can help with cognitive chunking, and let the user get a basic flow before trying to absorb a lot of the details.

src/mss/base.py
```diff
@@ -291,37 +296,47 @@ def close(self) -> None:
...
+            grab_region = region.as_region()
+        elif isinstance(region, Region):
+            grab_region = copy(region)
+        elif isinstance(region, dict):
+            grab_region = Region(
+                left=region["left"],
+                top=region["top"],
+                width=region["width"],
+                height=region["height"],
+            )
```
jholveck, line 326: May want to do type coercion here.

```suggestion
        elif isinstance(region, dict):
            grab_region = Region(
                left=int(region["left"]),
                top=int(region["top"]),
                width=int(region["width"]),
                height=int(region["height"]),
            )
```

jholveck (approved):
I think it's good as-is.  Unlike the previous revision, I don't have any major structural issues here.

I would suggest we commit it as it stands, but I do have several suggestions.  Many of these are documentation suggestions.  Many others are for code that's just near the edits, and not really part of @halldorfannar 's proposed changes, but I just happened to notice them.  I've put in suggestions for most of them, but most of these haven't been tested, or even run them through `check.sh`.

As I said, I suggest we commit this as it stands, but also consider these other suggestions for follow-up commits, or to split into new issues.

src/tests/test_save.py
BoboTiG: Looking at that commit makes me nostalgic haha.

I do not remember why I did a generator here, maybe was I thinking about memory resources optimization.

Anyway, we can change it, yes!

BoboTiG:
> would suggest we commit it as it stands, but I do have several suggestions. Many of these are documentation suggestions. Many others are for code that's just near the edits, and not really part of @halldorfannar 's proposed changes, but I just happened to notice them. I've put in suggestions for most of them, but most of these haven't been tested, or even run them through `check.sh`.

Agreed. I merge, feel free to open a follow up PR.

BoboTiG:
Thank you @halldorfannar 🥂

PR #569, opened 2026-08-13
Fix X-style geometry parsing

jholveck:

In #562, I had misinterpreted how negative numbers in X geometry strings are handled.  As a result of this, I also broke the ability to pass negative top and left coordinates, which are valid on Windows and macOS.

This fixes the negative numbers to be consistent with XParseGeometry, and restores the ability to have negative top and left coordinates in both styles.

- [X] Tests added/updated
- [X] Documentation updated
- [X] `./check.sh` passed

### AI assistance disclosure

- [X] No AI assistance was used to generate this contribution.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -297,9 +297,7 @@ Or via direct call from Python::
     -h, --help            show this help message and exit
     -c, --coordinates COORDINATES
                           the part of the screen to capture:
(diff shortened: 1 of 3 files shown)


jholveck:
This will need to be updated to conform to #566 , so it's not yet ready to merge.

jholveck:
I've updated it to use #566 's modern region handling.

PR #571, opened 2026-08-16
Delete temporary .png files created by tests

jholveck:
The previous behavior would remove any `*.png` or `*.png.old` files in the current directory, including those that were not created by the tests.  This might include files that the developer has for other purposes.  The previous behavior also only did this before running tests, not after.

This commit modifies the cleanup code to only remove new `.png` (and `.png.old`) files that were created by the tests, and to do so immediately after the tests have run.

In the future, we would probably be better off creating our PNG files in a directory managed by the [PyTest tmp_path fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html).  To that end, this change also is a step towards reliably identifying which tests are creating PNG files.

- [X] Tests added/updated
- [X] `./check.sh` passed

### AI assistance disclosure

- [ ] No AI assistance was used to generate this contribution. - AI autocomplete generated some completions that were used.

### DIFF
--- a/src/tests/conftest.py
+++ b/src/tests/conftest.py
@@ -2,8 +2,10 @@
 Source: https://github.com/BoboTiG/python-mss.
 """
 
+import operator
 import os
 from collections.abc import Callable, Generator
+from functools import reduce
 from hashlib import sha256
 from pathlib import Path
 from platform import system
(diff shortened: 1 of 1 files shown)


BoboTiG:
Yes, that would be better this way!

jholveck:
Ok, I just wanted to make sure you liked that direction.

PR #581, opened 2026-08-28
Improve the PR template

halldorfannar:
There are a few goals with this change, and I list them below.  Please suggest tweaks or better wording. English is my third language 😄 

* Make the PR template simpler 
   - remove the heading because every PR has a Title that is outside this template
   - instead of the "open to interpretation" `(...)` we now have a comment that tells contributors what we expect
* Make the template reflect our new reality
  - We now have release notes instead of a changelog.  We probably need more docs about how contributors are expected to update the release notes.  Maybe those guides exist and I just couldn't find them.
* Drive contributors to take action on all items 
  - before they would only check if a test was added/update, etc but this makes the GitHub interface look odd because a PR will show as having incomplete tasks.
  - similarly, with the changed wording for AI disclosure we can tell if people took action there or not.  If someone doesn't disclose we will assume AI was at least partially involved.

- [x] Tests added/updated (or your review concluded: not needed)
- [x] Documentation updated (or your review concluded: not needed)
- [x] Release notes added (or your review concluded: not needed)
- [x] `./check.sh` passed

### AI assistance disclosure

- AI was not used to create this change.


halldorfannar:
Created this from the wrong base branch. Will close and create another PR.

